PR #75, opened 2020-02-25
Remove safety section from README.

mbrubeck:
There are still some uses of `unsafe` in this crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,6 @@
 $ cargo add http-types
 ```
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@mbrubeck Hey, thanks for reporting! -- We should probably update the message to explain where we're using unsafe, rather than remove all information entirely. I'll write a quick patch for that. Thanks!

Issue #67, opened 2020-02-01
Implement Eq and PartialEq for HeaderName / HeaderValue and string types


yoshuawuyts:
We should enable the following:

```rust
assert_eq!(req.header("name"), Some("value"));
```

PR #80, opened 2020-03-02
Add equality checks between headers and strings

yoshuawuyts:
Implements https://github.com/http-rs/http-types/issues/67. Thanks!

### DIFF
--- a/src/headers/header_name.rs
+++ b/src/headers/header_name.rs
@@ -71,6 +71,30 @@ impl<'a> std::convert::TryFrom<&'a str> for HeaderName {
     }
 }
 
+impl PartialEq<str> for HeaderName {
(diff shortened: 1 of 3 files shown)


src/headers/header_name.rs
```diff
@@ -71,6 +71,30 @@ impl<'a> std::convert::TryFrom<&'a str> for HeaderName {
     }
 }
 
+impl PartialEq<str> for HeaderName {
+    fn eq(&self, other: &str) -> bool {
+        self.0 == other
```
kbillings: This is incorrect, header name comparisons must be case-insensitive.

yoshuawuyts: Oh yeah, good catch. We already normalize when casting from a string to a headername. But we should be doing the same for this comparison.

PR #83, opened 2020-03-08
Don't move MIME out of the body

tirr-c:
This PR fixes the panic when the same `Body` is reused on requests or responses without `Content-Type`.


tirr-c:
This actually fixes a bug and tests look okay, I'm going to self-merge this 😅

PR #90, opened 2020-03-31
impl std::error::Error for Error

stjepang:


yoshuawuyts:
We talked about this over chat a while back, and this patch is not the right fit for where we're taking this crate. Closing.

PR #92, opened 2020-04-06
Satisfy Clippy where ever possible

Koxiaet:
Changes include:
- Adding `is_empty` methods as well as `len` methods.
- Removing explicit lifetimes
- Adding `Default` impls for `Headers` and `Trailers`
- A few other minor changes that Clippy suggested.

I did not follow all the lints and make breaking changes.

### DIFF
--- a/src/headers/mod.rs
+++ b/src/headers/mod.rs
@@ -95,7 +95,7 @@ impl Headers {
     }
 
     /// An iterator visiting all header pairs in arbitrary order.
-    pub fn iter<'a>(&'a self) -> Iter<'a> {
+    pub fn iter(&self) -> Iter<'_> {
         Iter {
             inner: self.headers.iter(),
         }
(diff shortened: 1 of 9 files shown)


src/trailers.rs
```diff
@@ -131,27 +131,33 @@ impl Trailers {
...
+    pub fn values(&self) -> Values<'_> {
         self.headers.values()
     }
 }
 
+impl Default for Trailers {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
yoshuawuyts, line 159: Trailers should not be able to be constructed by hand; it's intended to only be constructed through the `trailers` method.

src/headers/mod.rs
```diff
@@ -95,33 +95,39 @@ impl Headers {
...
         Values::new(self.headers.values())
     }
 }
 
+impl Default for Headers {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
yoshuawuyts, line 130: `Headers` should never be constructed by hand, it should only ever be constructed by `Request`, `Response`, or `Trailers`.

PR #94, opened 2020-04-11
Add Client and Server traits

yoshuawuyts:
Adds `Client` and `Server` traits, replacing `http-service` and `http-client` and providing a bridge between the two traits.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,21 +115,25 @@ pub mod headers;
 pub mod mime;
 
 mod body;
+mod client;
 mod error;
 mod macros;
 mod method;
 mod request;
 mod response;
+mod server;
 mod status;
 mod status_code;
 mod type_map;
 mod version;
 
 pub use body::Body;
+pub use client::Client;
 pub use error::{Error, Result};
 pub use method::Method;
 pub use request::Request;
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -111,29 +112,36 @@ pub mod url {
...
 mod method;
 mod request;
 mod response;
-mod server;
 mod status;
 mod status_code;
 mod type_map;
 mod version;
 
+cfg_unstable! {
```
otavio, line 132: Is this macro worth it? You use it only once.

src/lib.rs
```diff
@@ -95,8 +95,9 @@
 
 #![forbid(rust_2018_idioms)]
 #![deny(missing_debug_implementations, nonstandard_style)]
-#![warn(missing_docs, missing_doc_code_examples, unreachable_pub)]
```
otavio, line 98: Why remove this? better would add an example for it.

yoshuawuyts: That'd be several hundred lines, which is out of scope for this PR.

yoshuawuyts:
Merging as this is marked "unstable"; allows us to experiment with it. Thanks!

Issue #89, opened 2020-03-26
impl PartialEq for Vec<HeaderValue>> and 'str, str, String


yoshuawuyts:
Getting a value from `Headers` and comparing it to a string is quite involved. The current way to do it is:

```rust
if let Some(encoding) = res.header(&TRANSFER_ENCODING) {
    if !encoding.is_empty() && encoding.last().unwrap().as_str() == "chunked" {
       // proceed
    }
}
```

Even though we implemented more convenient comparisons in #80, this is still quite involved. What we'd like to be able to do is:

```rust
if res.header(&TRANSFER_ENCODING) == "chunked" {
    // proceed
}
```

## Note on `Option<Vec<HeaderValue>>`

We currently return `Option<Vec<HeaderValue>>` from various of our methods. This is because HTTP by design can have multiple values for a single key. However there is some overlap between a vec of 0 elements, and a `None` value. The idea we've been talking about offline is to perhaps introduce a new type: `HeaderValues` which implements `Iterator<Item = HeaderValue>>` and guarantees to never be empty.

This would provide a convenient location to implement equality checks. Examples include:

- `PartialEq<[str]> for HeaderValues`
- `PartialEq<str> for HeaderValues`
- `PartialEq<String> for HeaderValues`
- `PartialEq<HeaderValue> for HeaderValues`

Returning HeaderValues from existing methods would be a breaking change, so we shouldn't do this right away. But this provides a cleaner path forward than implementing equality for `Vec<HeaderValue>`.

yoshuawuyts:
Orphan rules don't allow for this. We need to introduce `HeaderValues` for this to work. This is a breaking change, but also the only way forward it seems. Welp.

PR #98, opened 2020-04-16
add security submodule

yoshuawuyts:
One of the things we've talked about repeatedly during the development of http-types is the ability to encode a shared interface for all things HTTP. So far most of what we've introduced has been foundational components, creating a framework for which we can reason about HTTP. But most of that is now implemented, and we can start implementing many of the finer detailed of HTTP.

This patch introduces a set of typed security header constructors as part of a new `security` submodule. This has been adapted from [`http-rs/armor`](https://docs.rs/armor/1.2.0/armor/), which in turn is based on [helmetjs](https://github.com/helmetjs/helmet) with additional support for [CSP](https://github.com/http-rs/armor/pull/3). Most HTTP servers should probably be using this, but many aren't. This helps both lower the barrier, and educate people on what to use. Thanks!

__edit:__ Worth noting that some of armor's test remain commented out because creating the assertions wasn't very readable. We should enable these once #89 is implemented.

## Screenshot

Screenshot_2020-04-16 http_types security - Rust

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,18 +16,19 @@ default = []
 hyperium_http = ["http"]
 
 [dependencies]
-anyhow = "1.0.26"
-cookie = "0.12.0"
-infer = "0.1.2"
-omnom = "2.1.1"
-pin-project-lite = "0.1.0"
-url = "2.1.0"
 
 # Note(yoshuawuyts): used for async_std's `channel` only; use "core" once possible.
 async-std = { version = "1.4.0", features = ["unstable"] }
 
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
+
+anyhow = "1.0.26"
+cookie = "0.12.0"
+infer = "0.1.2"
+omnom = "2.1.1"
+pin-project-lite = "0.1.0"
+url = "2.1.0"
 
 [dev-dependencies]
 http = "0.2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,8 @@ infer = "0.1.2"
 omnom = "2.1.1"
 pin-project-lite = "0.1.0"
 url = "2.1.0"
+serde_json = "1.0.51"
+serde = { version = "1.0", features = ["derive"] }
 
 [dev-dependencies]
 http = "0.2.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,6 @@
 //! differently. But as a rule: if you know the size of the body, it's usually more efficient to
 //! declare it up front. But if you don't, things will still work.
 
-#![forbid(rust_2018_idioms)]
 #![deny(missing_debug_implementations, nonstandard_style)]
 #![warn(missing_docs, missing_doc_code_examples, unreachable_pub)]
 #![cfg_attr(test, deny(warnings))]
@@ -151,6 +150,8 @@ pub use crate::cookies::Cookie;
 
 #[doc(inline)]
 pub mod trailers;
+
+pub mod security;
 
 #[cfg(feature = "hyperium_http")]
 mod hyperium_http;
--- a/src/security/csp.rs
+++ b/src/security/csp.rs
@@ -1,0 +1,356 @@
+use crate::Headers;
+use serde::Serialize;
+use std::collections::HashMap;
+use std::fmt;
+
+/// Define source value
+///
+/// [read more](https://content-security-policy.com)
+#[derive(Debug)]
+pub enum Source {
+    /// Set source `'self'`
+    SameOrigin,
+    /// Set source `'src'`
+    SRC,
+    /// Set source `'none'`
+    None,
+    /// Set source `'unsafe-inline'`
+    UnsafeInline,
+    /// Set source `data:`
+    Data,
+    /// Set source `mediastream:`
+    Mediastream,
+    /// Set source `https:`
+    HTTPS,
+    /// Set source `blob:`
+    Blob,
+    /// Set source `filesystem:`
+    Filesystem,
+    /// Set source `'strict-dynamic'`
+    StrictDynamic,
+    /// Set source `'unsafe-eval'`
+    UnsafeEval,
+    /// Set source `*`
+    Wildcard,
+}
+
+impl fmt::Display for Source {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Source::SameOrigin => write!(f, "'self'"),
+            Source::SRC => write!(f, "'src'"),
+            Source::None => write!(f, "'none'"),
+            Source::UnsafeInline => write!(f, "'unsafe-inline'"),
+            Source::Data => write!(f, "data:"),
+            Source::Mediastream => write!(f, "mediastream:"),
+            Source::HTTPS => write!(f, "https:"),
+            Source::Blob => write!(f, "blob:"),
+            Source::Filesystem => write!(f, "filesystem:"),
+            Source::StrictDynamic => write!(f, "'strict-dynamic'"),
+            Source::UnsafeEval => write!(f, "'unsafe-eval'"),
+            Source::Wildcard => write!(f, "*"),
+        }
+    }
+}
+
+impl AsRef<str> for Source {
+    fn as_ref(&self) -> &str {
+        match *self {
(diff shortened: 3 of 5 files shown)


src/security/csp.rs
```diff
@@ -0,0 +1,367 @@
...
+/// security::default(&mut res);
+/// policy.apply(&mut res);
+///
+/// let name =
+///     headers::HeaderName::from_ascii("content-security-policy".to_owned().into_bytes()).unwrap();
+/// let headers = res.header(&name).unwrap();
+/// let header = headers.iter().next().unwrap();
+/// assert_eq!(header, "base-uri 'none'; default-src 'self' areweasyncyet.rs; object-src 'none'; script-src 'self' 'unsafe-inline'; upgrade-insecure-requests");
+/// ```
+
```
phansch, line 130: extra newline here

src/security/csp.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn plugin_types<T: AsRef<str>>(&mut self, source: T) -> &mut Self {
+        self.insert_directive("plugin-types", source);
+        self
+    }
+
+    /// Defines the Content-Security-Policy `require-sri-for` directive
+    ///
+    /// [MDN | require-sri-for](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-sri-for)
+    pub fn require_sri_for<T: AsRef<str>>(&mut self, source: T) -> &mut Self {
+        self.insert_directive("require-sri-for ", source);
```
phansch, line 268: Not sure if that whitespace was intentional?

```suggestion
        self.insert_directive("require-sri-for", source);
```

yoshuawuyts: I think this was indeed unintentional. Thanks!

src/security/csp.rs
yoshuawuyts: ah good catch; thanks!

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Set source `'unsafe-eval'`
+    UnsafeEval,
+    /// Set source `*`
+    Wildcard,
+}
+
+impl fmt::Display for Source {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Source::SameOrigin => write!(f, "'self'"),
```
dignifiedquire, line 40: why not just call as_ref() to avoid these getting out of sync?

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+
+/// Define source value
+///
+/// [read more](https://content-security-policy.com)
+#[derive(Debug)]
+pub enum Source {
+    /// Set source `'self'`
+    SameOrigin,
+    /// Set source `'src'`
+    SRC,
```
dignifiedquire, line 14: please change to Src

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Set source `'none'`
+    None,
+    /// Set source `'unsafe-inline'`
+    UnsafeInline,
+    /// Set source `data:`
+    Data,
+    /// Set source `mediastream:`
+    Mediastream,
+    /// Set source `https:`
+    HTTPS,
```
dignifiedquire, line 24: Https

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+            Source::UnsafeEval => "'unsafe-eval'",
+            Source::Wildcard => "*",
+        }
+    }
+}
+
+/// Define `report-to` directive value
+///
+/// [MDN | report-to](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to)
+#[derive(Serialize, Debug)]
```
dignifiedquire, line 78: why not Deserialize? Also please Clone, PartialEq and Eq

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    include_subdomains: Option<bool>,
+}
+
+/// Define `endpoints` for `report-to` directive value
+///
+/// [MDN | report-to](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to)
+#[derive(Serialize, Debug)]
+pub struct ReportToEndpoint {
+    url: String,
```
dignifiedquire, line 93: why not Deserialize? Also please Clone, PartialEq and Eq

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
...
+/// policy.apply(&mut res);
+///
+/// let name =
+///     headers::HeaderName::from_ascii("content-security-policy".to_owned().into_bytes()).unwrap();
+/// let headers = res.header(&name).unwrap();
+/// let header = headers.iter().next().unwrap();
+/// assert_eq!(header, "base-uri 'none'; default-src 'self' areweasyncyet.rs; object-src 'none'; script-src 'self' 'unsafe-inline'; upgrade-insecure-requests");
+/// ```
+#[derive(Debug)]
+pub struct ContentSecurityPolicy {
```
dignifiedquire, line 131: please Clone, PartialEq and Eq

src/security/csp.rs
```diff
@@ -0,0 +1,366 @@
+use crate::Headers;
+use serde::Serialize;
+use std::collections::HashMap;
+use std::fmt;
+
+/// Define source value
+///
+/// [read more](https://content-security-policy.com)
+#[derive(Debug)]
+pub enum Source {
```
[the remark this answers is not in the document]
dignifiedquire: maybe also Hash, to allow storing in hashmaps

src/security/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+pub fn dns_prefetch_control(mut headers: impl AsMut<Headers>) {
+    headers
+        .as_mut()
+        .insert("X-DNS-Prefetch-Control", "on")
+        .unwrap();
+}
+
+/// Set the frameguard level.
+#[derive(Debug, Clone)]
+pub enum FrameOptions {
```
dignifiedquire, line 61: Copy, PartialEq and Eq

src/security/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+pub fn xss_filter(mut headers: impl AsMut<Headers>) {
+    headers
+        .as_mut()
+        .insert("X-XSS-Protection", "1; mode=block")
+        .unwrap();
+}
+
+/// Set the Referrer-Policy level
+#[derive(Debug, Clone)]
+pub enum ReferrerOptions {
```
dignifiedquire, line 182: see above

src/lib.rs
```diff
@@ -93,7 +93,6 @@
 //! differently. But as a rule: if you know the size of the body, it's usually more efficient to
 //! declare it up front. But if you don't, things will still work.
 
-#![forbid(rust_2018_idioms)]
```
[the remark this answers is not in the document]
yoshuawuyts: We also deny warnings so that's the same as having it be an error. We also can't actually fix the code since Serde's codegen would need to be adjusted. I don't think there's a way out of this one unfortunately.

Cargo.toml
```diff
@@ -16,18 +16,21 @@ default = []
...
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
 
+anyhow = "1.0.26"
+cookie = "0.12.0"
+infer = "0.1.2"
+omnom = "2.1.1"
+pin-project-lite = "0.1.0"
+url = "2.1.0"
+serde_json = "1.0.51"
```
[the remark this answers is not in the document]
yoshuawuyts: We'd need to roll the internal derive by hand, which I'm not overly keen to do. We probably want to have json / form support in http-types at some point too, so I don't think there's really escaping Serde. Even embedded protocols such as [postcard](https://docs.rs/postcard/0.5.0/postcard/) work with Serde, so I don't think this particularly limits anyone.

src/security/csp.rs
yoshuawuyts: These need to be encoded *specifically* like this. Relying on a Debug impl wouldn't work here because for example `Source::Wildcard` needs to be encoded as `'*'`. Better to keep all of them consistent.

PR #100, opened 2020-04-21
Add conversions to and from bytes

phansch:
This adds `Body::from_bytes`, `Body.into_bytes`, `Response.body_bytes` and `Request.body_bytes` to make it easier for users to work with bytes directly.

cc https://github.com/http-rs/tide/pull/436#issuecomment-616435815

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,3 +1,4 @@
+use async_std::io::prelude::*;
 use async_std::io::{self, BufRead, Read};
 
 use std::fmt::{self, Debug};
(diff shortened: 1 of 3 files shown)


src/body.rs
```diff
@@ -109,6 +109,36 @@ impl Body {
...
+    /// use async_std::io::Cursor;
+    ///
+    /// let mut req = Response::new(StatusCode::Ok);
+    ///
+    /// let input = vec![1, 2, 3];
+    /// let len = input.len();
+    /// req.set_body(Body::from_bytes(input, Some(len)));
+    /// ```
+    pub fn from_bytes(
+        bytes: Vec<u8>,
```
phansch, line 132: So.. I guess ideally this would be `&[u8]`? However, that conflicts with the `'static` requirement of `reader: Box<dyn BufRead + Unpin + Send + Sync + 'static>`. I'm not really _why_ `Vec` doesn't conflict with that requirement either?

yoshuawuyts: This should probably take a `Vec<u8>` since it needs to keep ownership.

phansch:
I think I addressed all the comments so far and added `Body.into_bytes`, `Request.body_bytes` and `Response.body_bytes`.

PR #101, opened 2020-04-21
Body from

yoshuawuyts:


jbr:
These look quite convenient!
Would it make sense / be possible to add a generic `From<B> for Response where B: Into<Body>` (or whatever the orphan-rule-allowed version of that is) that creates a response with an ok status code and sets the Into<Body> as the body?

yoshuawuyts:
@jbr  done!

src/response.rs
```diff
@@ -491,32 +491,26 @@ impl AsMut<Headers> for Response {
...
-    fn from(b: Vec<u8>) -> Self {
+impl<T> From<T> for Response
+where
+    T: Into<Body>,
+{
+    fn from(value: T) -> Self {
+        let body: Body = value.into();
         let mut res = Response::new(StatusCode::Ok);
-        res.set_body(b);
+        res.set_body(body);
```
jbr, line 513: This is a minor tweak, but this could be 

```rust
impl<T> From<T> for Response
where
    T: Into<Body>,
{
    fn from(value: T) -> Self {
        let mut res = Response::new(StatusCode::Ok);
        res.set_body(value);
        res
    }
}
```

since replace_body does the `.into::<Body>()`

meta: is this sort of review useful?

yoshuawuyts: @jbr thanks heaps for suggesting this; yeah this is very useful!

PR #102, opened 2020-04-21
Match Surf's Body conversions

yoshuawuyts:
A big driving force behind http-types was the desire to share repeated logic between Tide and Surf, and in the process establish common idioms for web protocols. So far we've covered headers and URLs, but bodies are part of that as well.

So far for `Body` we have `Body::from_string` and `Body::into_string` with the corresponding `Request::body_string` and `Response::body_string` methods. This adds a parallel path for that for JSON as well. This is the first step to bringing us on par with [surf's conversions](https://docs.rs/surf/1.0.3/surf/struct.Request.html#method.body).

## Future directions
After this patch we should probably fill out the rest of the chart:

| text/plain | octet-stream | JSON | www-form-encoding | mime detection |
|---|---|---|---|---|
| `Body::from_string` | `Body::from_bytes` | `Body::from_json` | `Body::from_form` | `Body::from_file` |
| `Body::into_string` | `Body::into_bytes` | `Body::into_json` | `Body::into_form` | tbd |
| `Request::body_string` | `Request::body_bytes` | `Request::body_json` | `Request::body_form` | tbd |
| `Response::body_string` | `Response::body_bytes` | `Response::body_json` | `Response::body_form` |  tbd |

This will make implementing these methods in both Tide and Surf significantly easier, making it so we only need to forward function calls. Thanks!

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -300,7 +300,7 @@ impl Response {
     ///
     /// let cat = Cat { name: String::from("chashu") };
     /// let mut res = Response::new(StatusCode::Ok);    
-    /// res.set_body(Body::from_json(cat)?);
+    /// res.set_body(Body::from_json(&cat)?);
     ///
     /// let cat: Cat = res.body_json().await?;
     /// assert_eq!(&cat.name, "chashu");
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
This is a sibling PR to https://github.com/http-rs/http-types/pull/100

jbr:
similar to #106, this seems like it could benefit from using `impl TryInto<Body>` wherever `impl Into<Body>` is currently accepted?

yoshuawuyts:
@jbr oh yeah that's much better. Updating the patch!

src/response.rs
```diff
@@ -243,17 +244,46 @@ impl Response {
...
     /// # Ok(()) }) }
     /// ```
     pub async fn body_string(self) -> io::Result<String> {
         self.body.into_string().await
     }
 
+    /// Read the body as JSON.
+    ///
+    /// This consumes the response. If you want to read the body without
+    /// consuming the request, consider using the `take_body` method and
```
phansch, line 261:
```suggestion
    /// consuming the response, consider using the `take_body` method and
```

yoshuawuyts:
finished implementing the full conversion chart. The method overview in `Body` looks quite nice now:

Screenshot_2020-04-25 http_types Body - Rust

PR #106, opened 2020-04-22
Add From<async_std::fs::File> for Body

jbr:
+ adds an async_std feature flag
+ implements `From<async_std::fs::File> for http_types::Body`, conditional on the async_std feature flag
+ adds a `http_types::Body::from_file` function that uses the From conversion, also behind the feature flag

Concerns:
I implemented From instead of TryFrom because Request and Response signatures expect `Into<Body>`, but using `expect` in library code seems sketchy. Because this and potentially other conversions into Body are fallible, changing Request/Response set_body to accept a `TryInto<Body>` and propagating potential errors up to library user code would remove the need for the `expect`

I started to implement `From<&Path>`/`From<PathBuf>` but that's even more likely to fail than reading metadata from an already-open file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,14 +16,16 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = []
+default = ["async_std"]
 docs = ["unstable"]
 unstable = []
 hyperium_http = ["http"]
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Merging the work done in this patch into #102, and continuing that patch by filling out the conversions. Thanks!

PR #113, opened 2020-04-23
Fix incorrect headervalues comparison

yoshuawuyts:
We weren't validating we're indeed only one item long when comparing HeaderValues with singular other items. If we aren't we should reject. Thanks!


yoshuawuyts:
This fixes a pretty serious bug. Merging!

Issue #116, opened 2020-04-24
rename TypeMap to Extensions


yoshuawuyts:
And the methods to `ext` and `ext_mut`. Terminology around "local state" that returns a `TypeMap` indexed by `TypeId` has been hard to follow.

Instead talking about "here's a way to store more values in request and response" feels like a better fit. This is similar to Hyper's http crate, but with a different explainer.

yoshuawuyts:
Done in #118

PR #118, opened 2020-04-24
Rename TypeMap to Extensions

yoshuawuyts:
Implements #116. Thanks!

## Screenshots

Screenshot_2020-04-24 http_types - Rust

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -7,26 +7,25 @@ use std::collections::HashMap;
 use std::fmt;
 use std::hash::{BuildHasherDefault, Hasher};
 
-/// Store and retrieve values by `TypeId`.
+/// A type to store extra data inside `Request` and `Response`.
 ///
-/// Map type that allows storing any `Sync + Send + 'static` type. Instances can
-/// be retrieved from [`Request::local`](struct.Request.html#method.local) +
-/// [`Response::local`](struct.Response.html#method.local) and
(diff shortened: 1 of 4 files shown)


Fishrock123:
As a newcomer, this seems more clear. :+1:

yoshuawuyts:
@Fishrock123 the issue I see with the word "local" is that the question becomes: "local to what?" `LocalStorage` may be scoped locally to the thread, runtime, framework, request, endpoint, or middleware. `Extensions` is less overloaded -- at worst it may not be clear *what* it extends, but the assumption will always be that it extends something in the crate -- and good docs can help explain what it does.

Fishrock123:
What about `ContinuationStorage` then?

yoshuawuyts:
@Fishrock123 if, like me, you don't know what a "continuation" is in this context, that name doesn't seem very intuitive. It's also fairly long. "Extensions" is shortened in the codebase as `Request::ext()` and `Response::ext()`. In contrast smth like `Request::cont_store` seems less intuitive to me, but also longer.

I feel overall "extensions" is the best fit we have. With the added benefit that other crates in the Rust ecosystem use the same name for the same concept as well.

Fishrock123:
A bit late, but `async_std` does also use the term "local storage" for tasks, and the api is "LocalKey": https://docs.rs/async-std/1.5.0/async_std/task/index.html#task-local-storage

PR #119, opened 2020-04-24
Add peer and local socket addrs, as well as Request::remote

jbr:
this is the first PR of three to address https://github.com/http-rs/tide/issues/462

This also adds an initial implementation of Request::remote and Request::forwarded_for. Parsing the Forwarded header value probably would be more elegant with a proper parsing or regex library, but I didn't want to add an additional dependency just for this.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -60,12 +60,12 @@ impl Request {
         }
     }
 
-    /// get the peer socket address for the underlying transport, if appropriate
+    /// Get the peer socket address for the underlying transport, if appropriate
     pub fn peer_addr(&self) -> Option<SocketAddr> {
         self.peer_addr
     }
 
-    /// get the local socket address for the underlying transport, if appropriate
+    /// Get the local socket address for the underlying transport, if appropriate
     pub fn local_addr(&self) -> Option<SocketAddr> {
         self.local_addr
     }
--- a/src/request.rs
+++ b/src/request.rs
@@ -3,7 +3,6 @@ use async_std::sync;
 
 use std::convert::TryInto;
 use std::mem;
-use std::net::SocketAddr;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
@@ -37,8 +36,8 @@ pin_project_lite::pin_project! {
         #[pin]
         body: Body,
         local: TypeMap,
-        local_addr: Option<SocketAddr>,
-        peer_addr: Option<SocketAddr>,
+        local_addr: Option<String>,
+        peer_addr: Option<String>,
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/request.rs
```diff
@@ -52,6 +54,52 @@ impl Request {
...
+    }
+
+    /// Get the remote address for this request.
+    pub fn remote(&self) -> Option<&str> {
+        self.forwarded_for().or(self.peer_addr())
+    }
+
+    /// Parses the Forwarded or X-Forwarded-For headers.
+    /// The returned String will either be an IP address or a domain and an optional port.
+    pub fn forwarded_for(&self) -> Option<&str> {
```
yoshuawuyts, line 89: This should not be part of the public API. The way to access values produced by this should be through `remote` or `host`.

```suggestion
    fn forwarded_for(&self) -> Option<&str> {
```

src/request.rs
```diff
@@ -52,6 +54,52 @@ impl Request {
...
+
+    #[doc(hidden)]
+    pub fn set_peer_addr(&mut self, peer_addr: Option<impl std::string::ToString>) {
+        self.peer_addr = peer_addr.map(|addr| addr.to_string());
+    }
+
+    #[doc(hidden)]
+    pub fn set_local_addr(&mut self, local_addr: Option<impl std::string::ToString>) {
+        self.local_addr = local_addr.map(|addr| addr.to_string());
+    }
```
yoshuawuyts, line 70: This should be public and documented. Tide / Surf shouldn't expose this method, but `http-types` should.

```suggestion
    pub fn set_local_addr(&mut self, local_addr: Option<impl std::string::ToString>) {
        self.local_addr = local_addr.map(|addr| addr.to_string());
    }
```

src/request.rs
```diff
@@ -52,6 +54,52 @@ impl Request {
...
             local: TypeMap::new(),
+            peer_addr: None,
+            local_addr: None,
+        }
+    }
+
+    #[doc(hidden)]
+    pub fn set_peer_addr(&mut self, peer_addr: Option<impl std::string::ToString>) {
+        self.peer_addr = peer_addr.map(|addr| addr.to_string());
+    }
```
yoshuawuyts, line 65: This should be public and documented. Tide / Surf shouldn't expose this method, but `http-types` should.

```suggestion
    pub fn set_peer_addr(&mut self, peer_addr: Option<impl std::string::ToString>) {
        self.peer_addr = peer_addr.map(|addr| addr.to_string());
    }
```

src/response.rs
```diff
@@ -331,6 +335,26 @@ impl Response {
         self.version
     }
 
+    #[doc(hidden)]
+    pub fn set_peer_addr(&mut self, peer_addr: Option<impl std::string::ToString>) {
+        self.peer_addr = peer_addr.map(|addr| addr.to_string());
+    }
```
yoshuawuyts, line 341: This should be public and documented. Tide / Surf shouldn't expose this method, but `http-types` should.

```suggestion
    pub fn set_peer_addr(&mut self, peer_addr: Option<impl std::string::ToString>) {
        self.peer_addr = peer_addr.map(|addr| addr.to_string());
    }
```

src/response.rs
```diff
@@ -331,6 +335,26 @@ impl Response {
...
+
+    #[doc(hidden)]
+    pub fn set_local_addr(&mut self, local_addr: Option<impl std::string::ToString>) {
+        self.local_addr = local_addr.map(|addr| addr.to_string());
+    }
```
yoshuawuyts, line 346: This should be public and documented. Tide / Surf shouldn't expose this method, but `http-types` should.

```suggestion
    pub fn set_local_addr(&mut self, local_addr: Option<impl std::string::ToString>) {
        self.local_addr = local_addr.map(|addr| addr.to_string());
    }
```

jbr:
Addressed all changes, thanks for the review!

PR #127, opened 2020-05-13
Improve usage ergonomics

yoshuawuyts:
I need to write some more words on this, but in short this patch exists to make this crate much nicer to work with. In particular constructing new responses and comparing headers was pretty tricky. This makes that all much nicer to do.

## Changes
- this enables shorthand constructors for req/res by taking values by `impl TryInto`.
    - this still allows all existing code to work -- e.g. constructing req/res based on user input can still be done, and errors can still be handled.
    - in general this removes quite a few unwraps, bringing it overall closer to ergonomics in other langs.

### DIFF
--- a/src/headers/header_name.rs
+++ b/src/headers/header_name.rs
@@ -71,7 +71,7 @@ impl From<&HeaderName> for HeaderName {
 
 impl<'a> From<&'a str> for HeaderName {
     fn from(value: &'a str) -> Self {
-        Self::from_str(value).unwrap()
+        Self::from_str(value).expect("String slice should be valid ASCII")
     }
 }
 
--- a/src/headers/header_name.rs
+++ b/src/headers/header_name.rs
@@ -63,11 +63,9 @@ impl FromStr for HeaderName {
     }
 }
 
-impl<'a> std::convert::TryFrom<&'a str> for HeaderName {
-    type Error = Error;
-
-    fn try_from(value: &'a str) -> Result<Self, Self::Error> {
-        Self::from_str(value)
+impl<'a> From<&'a str> for HeaderName {
+    fn from(value: &'a str) -> Self {
+        Self::from_str(value).unwrap()
     }
 }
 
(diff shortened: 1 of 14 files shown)


jbr:
This seems like a massive improvement :thumbsup:

jbr:
Should the status arg to `Response::with_reader<R>(status: u16, reader: R)` also become a generic TryInto<StatusCode> to line up with the changes to `Response::new`?

jbr:
Previously `req.header(&HEADER_NAME)` was allowed so there is likely code out there like that.  Would it be very unidiomatic to add
```rust
impl<'a> From<&'a HeaderName> for HeaderName {
    fn from(value: &'a HeaderName) -> HeaderName {
        value.clone()
    }
}
```
so that `req.header(&WHATEVER)` will continue to work?

yoshuawuyts:
@jbr oh it looks like we no longer have `Response::with_reader` -- didn't realize, but that's good!

@jbr regarding passing headers by reference -- I see clear reasons why it wouldn't make sense, but then again: this would make it closer to the behavior of `HashMap` in std, so sure, let's add it (:

src/headers/header_name.rs
```diff
@@ -63,11 +63,9 @@ impl FromStr for HeaderName {
...
 }
 
-impl<'a> std::convert::TryFrom<&'a str> for HeaderName {
-    type Error = Error;
-
-    fn try_from(value: &'a str) -> Result<Self, Self::Error> {
-        Self::from_str(value)
+impl<'a> From<&'a str> for HeaderName {
+    fn from(value: &'a str) -> Self {
+        Self::from_str(value).unwrap()
```
rylev, line 68: expect with message instead of unwrap?

yoshuawuyts:
@rylev yeah, this will have to be a major version update.

src/headers/header_name.rs
yoshuawuyts: good call!

PR #131, opened 2020-05-15
Request::host

### LINKED ISSUE #129: Add Request::host
After https://github.com/http-rs/http-types/pull/119 we're still missing `Response::host`, the only method left from https://github.com/http-rs/http-types/issues/117. This should be modeled after https://docs.rs/actix-web/3.0.0-alpha.2/actix_web/dev/struct.ConnectionInfo.html#method.host. Thanks!

cc/ @jbr

jbr:
closes #129

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -93,25 +93,48 @@ impl Request {
     }
 
     /// Get the remote address for this request.
+    /// This is determined in the following priority:
+    /// 1. `Forwarded` header `for` key
+    /// 2. The first `X-Forwarded-For` header
+    /// 3. Peer address of the transport
     pub fn remote(&self) -> Option<&str> {
-        self.forwarded_for().or(self.peer_addr())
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -68,9 +69,23 @@ impl Response {
...
             local_addr: None,
+            host: None,
         }
     }
 
+    /// Sets the host that this response is from, using
+    /// [Request::host](crate::Request::host)
+    pub fn set_host_from_request(&mut self, request: &crate::Request) {
+        self.host = request.host().map(|h| h.to_string())
+    }
```
yoshuawuyts, line 80: I think having a `Response::set_host` method might be a better fit; it should be about the same length to copy it from a request, but much easier:

```rust
let mut req = ...;
let mut res = ...;

res.set_host(req.host());

jbr:
This makes a lot more sense! Reverted the changes to Response, should be good to go

PR #132, opened 2020-05-16
Add mime sniffing to Body::from_file

yoshuawuyts:
Adds mime sniffing to `Body::from_file`. This first tries magic bytes, then extensions, and finally falls back to octet streams. This is required for https://github.com/http-rs/tide/pull/461 to continue to work when we switch Tide over to an API that more closely resembles http-types. In addition this PR adds a few more common mime type constants that people might want to use. And it fixes some equality checks and missing methods. Thanks!


src/body.rs
```diff
@@ -448,3 +462,33 @@ impl BufRead for Body {
...
+#[cfg(feature = "async_std")]
+fn guess_ext(path: &Path) -> Option<Mime> {
+    let ext = path.extension().map(|p| p.to_str()).flatten();
+    match ext {
+        Some("html") => Some(mime::HTML),
+        Some("js") | Some("mjs") => Some(mime::JAVASCRIPT),
+        Some("json") => Some(mime::JSON),
+        Some("css") => Some(mime::CSS),
+        Some("svg") => Some(mime::SVG),
+        Some("wasm") => Some(mime::WASM),
```
jbr, line 491: This shouldn't be necessary, as infer supports WASM detection: https://github.com/bojand/infer#application

src/body.rs
```diff
@@ -448,3 +462,33 @@ impl BufRead for Body {
...
+}
+
+/// Look at the extension of a file to determine the mime type.
+/// This is useful for plain-text formats such as HTML and CSS.
+#[cfg(feature = "async_std")]
+fn guess_ext(path: &Path) -> Option<Mime> {
+    let ext = path.extension().map(|p| p.to_str()).flatten();
+    match ext {
+        Some("html") => Some(mime::HTML),
+        Some("js") | Some("mjs") => Some(mime::JAVASCRIPT),
```
jbr, line 487: Maybe `jsonp` as well for js? Do people still use that as a cors hack?

src/body.rs
```diff
@@ -448,3 +462,33 @@ impl BufRead for Body {
         Pin::new(&mut self.reader).consume(amt)
     }
 }
+
+/// Look at first few bytes of a file to determine the mime type.
+/// This is used for various binary formats such as images and videos.
+#[cfg(feature = "async_std")]
+async fn peek_mime(file: &mut async_std::fs::File) -> io::Result<Option<Mime>> {
+    // Reading the first 8 bytes should be enough to sniff the mime type.
```
jbr, line 470: Looking at infer's code, it seems like quite a few of the matchers go beyond 8 bytes. Is there a meaningful performance penalty to providing the first 300 bytes or something like that to infer? I can imagine a later bug where someone is like "why isn't my [jpeg2000](https://github.com/bojand/infer/blob/master/src/matchers/image.rs#L23)|[wav](https://github.com/bojand/infer/blob/master/src/matchers/audio.rs#L45)|[epub](https://github.com/bojand/infer/blob/master/src/matchers/archive.rs#L8) file being served correctly, infer supports it?" and the bugfix is to change this number

300 suggested because although happily people usually gzip their tars, the tar format seems to have quite a preamble before it gets around to saying what it is: https://github.com/bojand/infer/blob/master/src/matchers/archive.rs#L48-L55

src/body.rs
yoshuawuyts: I don't know of anyone using jsonp still. But no harm in adding it.

src/body.rs
yoshuawuyts: ah, great catch!

src/body.rs
yoshuawuyts: heh, good point. updating it to 300 bytes!

src/body.rs
```diff
@@ -485,10 +485,9 @@ fn guess_ext(path: &Path) -> Option<Mime> {
     match ext {
         Some("html") => Some(mime::HTML),
         Some("js") | Some("mjs") => Some(mime::JAVASCRIPT),
-        Some("json") => Some(mime::JSON),
+        Some("json") | Some("jsonp") => Some(mime::JSON),
```
jbr, line 488: I think jsonp is actually "application/javascript" because it's of the form callback(data) https://blog.abelotech.com/posts/json-jsonp-correct-mime-type/

yoshuawuyts: ahhh, gotcha!

PR #141, opened 2020-05-21
Fix header append

yoshuawuyts:
Addresses some of the inconsistencies pointed out by https://github.com/http-rs/tide/pull/520. Thanks!


yoshuawuyts:
> Am I right that it's pretty much a matter of string encoding and should be very unlikely for anyone to run into?

Yes, accurate. That's the only case. We'll never panic however if a valid `HeaderValue` / `HeaderName` has been provided already, which provides a way to construct values with the errors handled.

PR #144, opened 2020-05-21
Remove the Sync bound for Body::from_reader

sunli829:
I think this reader does not need `Sync`, because it is impossible for multiple threads to read `Body` at the same time, what do you think?


yoshuawuyts:
@sunli829 thanks for opening this! One would indeed assume this is true, but we've had repeated issues in the past when Body wasn't Sync. For example: https://github.com/http-rs/surf/issues/97. We should keep this bound to prevent future cases like that.

PR #152, opened 2020-05-24
Add feature for enabling signed and encrypted cookie support

kyrias:
The cookie crate uses the ring crate's HMAC module to implement signed
cookies and its AEAD module for authenticated encrypted cookies.


kyrias:
(Turns out that cookie 0.14.0 will use the RustCrypto crates instead of ring, though the CookieJar interface appears to be unchanged.)

kyrias:
Sure, I can file another PR adding it to the default feature as well.

PR #153, opened 2020-05-24
Fix compilation on wasm32-unknown-unknown

Perseus101:
The most recent release of http-types, 2.0.1, fails to compile on the wasm32-unknown-unknown target. I haven't checked previous versions to see exactly when the bug was introduced, but I suspect it was when async_std was updated to 1.6.

The cause of the problem is that async_std::fs is not available on wasm32-unknown-uknown. This PR gates all usages of that module behind the correct compilation options.

Additionally, I needed the AsRef<str> implementation on Method when fixing wasm compilation for http-client.

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -92,3 +92,19 @@ impl<'a> std::convert::TryFrom<&'a str> for Method {
         Self::from_str(value)
     }
 }
+
+impl AsRef<str> for Method {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::Get => "GET",
+            Self::Head => "HEAD",
+            Self::Post => "POST",
+            Self::Put => "PUT",
+            Self::Delete => "DELETE",
+            Self::Connect => "CONNECT",
+            Self::Options => "OPTIONS",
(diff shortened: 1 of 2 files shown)


src/method.rs
```diff
@@ -92,3 +92,19 @@ impl<'a> std::convert::TryFrom<&'a str> for Method {
         Self::from_str(value)
     }
 }
+
+impl AsRef<str> for Method {
```
jbr, line 96: Is this required to wasm32-unknown-unknown?

Perseus101: Strictly, no. But I needed it when fixing both http-client(https://github.com/jbr/http-client/pull/1) and surf(I'll open a pr in a few minutes) compilation.

Perseus101: Surf pr: https://github.com/jbr/surf/pull/1

Perseus101: If adding this implementation is a concern, I can remove it and do something else downstream. This just made the most sense IMO for how method was being used downstream.

yoshuawuyts: No objections to this bound.

Issue #155, opened 2020-05-24
Status should take TryInto<StatusCode>


yoshuawuyts:
When writing code quickly it's nice to be able to quickly jot down statuscodes for errors. It'd be nice if the Status trait was able to accept a wider range of input.

```rust
// current
let bar = foo.status(StatusCode::NotFound)?;

// proposed
let bar = foo.status(404)?;
```

We already do this in the Response constructor as well. This should provide a template for how to go about it. Marking as a good first issue!

zeroed:
Hello.

Thank you for opening and labelling `good-first-issues`; I would be glad to work on this.

PR #157, opened 2020-05-25
Status should take TryInto<StatusCode>

zeroed:
Change the trait bound of `S` in the `Status` trait to accept a `TryInto<StatusCode>`; this would allow the `Status` trait to accept a wider range of input.

Uses the same approach as `Response::new`.

Refs: #155 

Draft PR until the checklist is complete:

- [X] feature implementation.
- [x] tests added.
- [x] documentation updated.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -656,8 +656,15 @@ impl<'a> IntoIterator for &'a mut Response {
 #[cfg(test)]
 mod test {
     use super::Response;
+
     #[test]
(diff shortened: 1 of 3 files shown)


zeroed:
The commit a2e58a1 contains only the linting from `rust fmt`. It can be kept here, moved in a different MR or simply dropped; I would leave this to the judgement of the maintainers.

Thanks

Issue #154, opened 2020-05-24
Request::query should match Tide's behavior


yoshuawuyts:
In https://github.com/http-rs/tide/pull/537 we couldn't update the `Request::query` method to `http_types`' version because tests were failing. We should port Tide's tests to `http-types` and fix the query method to match the behavior. This means correctly parsing empty query strings, and setting a 400 status code.

This should be fairly straight forward, so marking as a good first issue!

## Refs
- Tide tests: https://github.com/http-rs/tide/blob/master/tests/querystring.rs
- Request::query: https://docs.rs/http-types/2.0.1/http_types/struct.Request.html#method.query

plippe:
Hey;

Happy to give it a try.

plippe:
@yoshuawuyts would you mind releasing a new version of http-types with my changes. That would allow me to update tide.

Fishrock123:
@plippe We'll probably do that tomorrow or so - Yosh mentioned doing a tide 0.10.0 tomorrow, which we would want a new http-types release for.

In the mean time, you can make a test/dev branch by updating tide's `Cargo.toml` to:
```
http-types = { git = "https://github.com/http-rs/http-types.git" }
```

if you made a draft branch that would probably help us do the update!

PR #162, opened 2020-05-28
Move querystring tests from tide

plippe:
https://github.com/http-rs/http-types/issues/154

### DIFF
--- a/tests/querystring.rs
+++ b/tests/querystring.rs
@@ -1,0 +1,54 @@
+use http_types::{url::Url, Method};
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+struct Params {
+    msg: String,
+}
+
+#[derive(Deserialize)]
+struct OptionalParams {
+    _msg: Option<String>,
+    _time: Option<u64>,
+}
+
+#[test]
+fn successfully_deserialize_query() {
+    let req = http_types::Request::new(
+        Method::Get,
+        Url::parse("http://example.com/?msg=Hello").unwrap(),
+    );
+
+    let params = req.query::<Params>();
+    assert!(params.is_ok());
+    assert_eq!(params.unwrap().msg, "Hello");
+}
+
+#[test]
+fn unsuccessfully_deserialize_query() {
+    let req = http_types::Request::new(Method::Get, Url::parse("http://example.com/").unwrap());
+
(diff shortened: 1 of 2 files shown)


tests/querystring.rs
```diff
@@ -0,0 +1,54 @@
...
+    assert_eq!(params.unwrap().msg, "Hello");
+}
+
+#[test]
+fn unsuccessfully_deserialize_query() {
+    let req = http_types::Request::new(Method::Get, Url::parse("http://example.com/").unwrap());
+
+    let params = req.query::<Params>();
+    assert!(params.is_err());
+    assert_eq!(params.err().unwrap().to_string(), "invalid data");
```
plippe, line 33: Different error message to the one in [original test](https://github.com/http-rs/tide/blob/master/tests/querystring.rs#L64)

tests/querystring.rs
```diff
@@ -0,0 +1,54 @@
...
+        Url::parse("http://example.com/?error=should_fail").unwrap(),
+    );
+
+    let params = req.query::<Params>();
+    assert!(params.is_err());
+    assert_eq!(params.err().unwrap().to_string(), "missing field `msg`");
+}
+
+#[test]
+fn empty_query_string_for_struct_with_no_required_fields() {
```
plippe, line 49: Test fails

plippe:
The failing test and the different error message are related to the missing query string. The tests are the same once I add any query string to both: e.g. `http://example.com/?`.
This is due to the query function that [fails when no query is present](https://github.com/http-rs/http-types/blob/master/src/request.rs#L634). 

I am unsure of what would be preferred:
- Should I add a query string to all tests?
- Should the query function return an `Option<Result<T>>`?

plippe:
I also believe that the `query` function might be worth moving to the `Url` struct instead of the `Request`.

plippe:
Completely missed that "default to an empty string".
It should work now (famous last words).

Also wondering about moving the implementation to the `Url`. It feels wrong to need a request to extract a URL's query string.

yoshuawuyts:
> Also wondering about moving the implementation to the Url

It could be worth bringing up. However `url` ships with a decoder already, just not the `serde` bounds. It may just be they don't want the dependency, even if optional.

---

That said this patch looks great; let's merge it!

Issue #164, opened 2020-05-29
Consider using u64 for Body length


VShell:
Currently, the Body's length is usize - which implies that a 32-bit web server could only respond with 4GB of data without using chunked encoding, even if they're streaming data which could be much larger. This could be resolved by using a u64 instead.

yoshuawuyts:
Ah yeah, we should consider this change for `http-types@3`. [`fs::Metadata::len`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len) returns a u64 as well.

PR #167, opened 2020-05-30
Body: make len() and related use u64 rather than usize.

Fishrock123:
Breaking change.

Closes https://github.com/http-rs/http-types/issues/164


yoshuawuyts:
We should probably hold off on merging this until we feel ready for a 3.0.0 release.

joshtriplett:
Rebased via the command line, then pushed.

Thank you!

PR #174, opened 2020-06-05
Response: add Error storage, retrieval, conversion

Fishrock123:
This allows for robust creation of `Response`s directly from `Error`s, with error
capture for future reference, and retrieval via `error() -> Option<&Error>`.

Refs: #169
Refs: http-rs/tide#546
Refs: http-rs/tide#532
Refs: http-rs/tide#452

For posterity, see discussion in discord following: https://discordapp.com/channels/598880689856970762/649056551835009097/718227631421784084

-------

~~Optionally also commit 2 (feel free to merge without if commit 2 is controversial)~~ (https://github.com/http-rs/http-types/pull/174/commits/777f4637633941a4dc053718ffe1c5d560d4be29)

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -483,6 +483,11 @@ impl Response {
         self.error.take()
     }
 
+    /// Sets an `Error` on the response, accessible via `error()` and `take_error()`.
+    pub fn set_error(&mut self, error: Error) {
+        self.error = Some(error);
+    }
+
     /// Get the HTTP version, if one has been set.
     ///
     /// # Examples
--- a/src/response.rs
+++ b/src/response.rs
@@ -15,7 +15,7 @@ use crate::headers::{
 };
 use crate::mime::Mime;
 use crate::trailers::{self, Trailers};
-use crate::{Body, Extensions, StatusCode, Version};
+use crate::{Body, Error, Extensions, StatusCode, Version};
 
 cfg_unstable! {
     use crate::upgrade;
@@ -50,6 +50,7 @@ pin_project_lite::pin_project! {
         ext: Extensions,
         local_addr: Option<String>,
         peer_addr: Option<String>,
+        error: Option<Error>,
     }
 }
 
@@ -85,6 +86,7 @@ pin_project_lite::pin_project! {
         ext: Extensions,
         local_addr: Option<String>,
         peer_addr: Option<String>,
+        error: Option<Error>,
     }
 }
 
@@ -111,6 +113,7 @@ impl Response {
             ext: Extensions::new(),
             peer_addr: None,
             local_addr: None,
+            error: None,
         }
     }
 
@@ -140,6 +143,7 @@ impl Response {
             ext: Extensions::new(),
             peer_addr: None,
             local_addr: None,
+            error: None,
         }
     }
 
@@ -469,6 +473,16 @@ impl Response {
         self.body.is_empty()
     }
 
+    /// Returns an optional reference to the `Error` if the response was created from one, or else `None`.
+    pub fn error(&self) -> Option<&Error> {
+        self.error.as_ref()
+    }
+
+    /// Takes the `Error` from the response if one exists, replacing it with `None`.
+    pub fn take_error(&mut self) -> Option<Error> {
+        self.error.take()
+    }
+
     /// Get the HTTP version, if one has been set.
     ///
     /// # Examples
@@ -641,8 +655,10 @@ impl Response {
 }
 
 impl Clone for Response {
-    /// Clone the response, resolving the body to `Body::empty()` and removing
-    /// extensions.
+    /// Clone the response, with some exceptions:
+    /// - The body is resolved to `Body::empty()`.
+    /// - Any attached extensions are not cloned.
+    /// - Any attached error is not cloned.
     fn clone(&self) -> Self {
         Self {
             status: self.status.clone(),
@@ -661,6 +677,7 @@ impl Clone for Response {
             ext: Extensions::new(),
             peer_addr: self.peer_addr.clone(),
             local_addr: self.local_addr.clone(),
+            error: None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Fishrock123:
Backed out commit 2 after discussion: https://discordapp.com/channels/598880689856970762/649056551835009097/718265585602920468

src/response.rs
```diff
@@ -136,9 +139,93 @@ impl Response {
...
+            body,
+            trailers_sender: Some(trailers_sender),
+            trailers_receiver: Some(trailers_receiver),
+            ext: Extensions::new(),
+            peer_addr: None,
+            local_addr: None,
+            error: Some(error),
+        };
+        if !was_server_error {
+            res.set_content_type(mime::PLAIN);
```
jbr, line 181: Does the body content type not get transferred correctly?

src/response.rs
```diff
@@ -136,9 +139,93 @@ impl Response {
...
+    /// If the `Error`'s status had a status code that was not a 5XX server
+    /// error code, the response body will be set to the error's message, and
+    /// the content-type header will be set to `http_types::mime::Plain`.
+    #[cfg(not(feature = "unstable"))]
+    pub fn from_error(error: Error) -> Self {
+        // Only send the message if it is a non-500 range error. All
+        // errors default to 500 by default, so sending the error
+        // body is opt-in at the call site.
+        let was_server_error = error.status().is_server_error();
+        let body = if !was_server_error {
```
jbr, line 161: `if !predicate {/*false condition*/} else {/*true condition*/}` is more confusing to read than `if predicate { */true condition */ } else {/* false condition */}`

src/response.rs
```diff
@@ -650,6 +747,7 @@ impl Clone for Response {
             ext: Extensions::new(),
             peer_addr: self.peer_addr.clone(),
             local_addr: self.local_addr.clone(),
+            error: None,
```
jbr, line 750: Is this because we don't have a Clone bound on the error? This seems like a source of confusion

src/response.rs
```diff
@@ -136,9 +139,93 @@ impl Response {
...
+    /// error code, the response body will be set to the error's message, and
+    /// the content-type header will be set to `http_types::mime::Plain`.
+    #[cfg(not(feature = "unstable"))]
+    pub fn from_error(error: Error) -> Self {
+        // Only send the message if it is a non-500 range error. All
+        // errors default to 500 by default, so sending the error
+        // body is opt-in at the call site.
+        let was_server_error = error.status().is_server_error();
+        let body = if !was_server_error {
+            Body::from_string(error.to_string())
```
jbr, line 162: Why are we generating bodies at all here? Later middleware can generate a body if that's what the application needs, and a plaintext body isn't necessarily a particularly useful default. If we want a default response body for errors, I think it makes more sense to add a middleware that generates an error body based on the request's Accept header or something like that. I think this should just set the error and status

src/response.rs
Fishrock123: Yeah, I don't really know how to Clone `anyhow::Error` /  `std::error:Error` correctly.

src/response.rs
Fishrock123: This is taken from tide -- basically with hopes of consolidating this: https://github.com/http-rs/tide/pull/546/files#diff-3dbd16223f25e8b5e0bcf099743107e9R10-R23

Perhaps that's not quite the right move but it seemed fitting enough.

src/response.rs
Fishrock123: Ah looks like it does actually, tide just did this unnecessarily.

src/response.rs
jbr: I think this PR obviates that approach. If a middleware has not set a response body, the app doesn't need a response body

src/response.rs
Fishrock123: Err, nope. We don't use `set_body`, we are constructing the struct directly, so no logic is available. We could instead call `copy_content_type_from_body` but that seems like unnecessary overhead, since it will only ever be this anyways.

src/response.rs
jbr: After this change, #564 will just be a matter of the cors middleware setting a header on the response regardless of whether it has an Error, since https://github.com/http-rs/tide/pull/546/files#diff-dc3595a3d4ca6ef52648b5c158a0e100R164 will already be a Response

Fishrock123: Hmm, I'm not sure. Does this hurt? This allows to default case to be handled automatically if the user has not set a body or headers, and doesn't need some fancy dirty-checking in tide or surf to check if those should set these after middleware.

jbr: This would be set *before* the error handling middleware though, which would result in allocating an unused body. The error handling middleware would have to check if it's a default body or an app-set body. I don't see any advantage to having default body behavior, since almost every production app will need to replace it. If there's default behavior, I think it should be implemented in tide, as a default middleware. This seems like an unnecessary exception to an otherwise simple-to-explain pattern.

src/response.rs
Fishrock123: I guess I can go back to storing it in an `Rc` (or an `Arc`? Idk how much thread-safety is involved here) and then return a reference out of the `Rc` on `error()`...

jbr: That seems like a possibility, but I'm not sure who uses Response::clone or how important it is. I don't think tide or surf clone a response as part of normal behavior, so maybe it's just for testing?

src/response.rs
```diff
@@ -49,6 +50,7 @@ pin_project_lite::pin_project! {
         ext: Extensions,
         local_addr: Option<String>,
         peer_addr: Option<String>,
+        error: Option<Rc<Error>>,
```
Fishrock123, line 53: Note: may need to be an `Arc`? I am not sure.

Fishrock123: Nvm, Tide needs an `Arc`.

src/response.rs
```diff
@@ -465,6 +524,16 @@ impl Response {
         self.body.is_empty()
     }
 
+    /// Returns an optional reference to the `Error` if the response was created from one, or else `None`.
+    pub fn error(&self) -> Option<&Error> {
+        self.error.as_ref().map(|v| v.as_ref())
+    }
+
+    /// Takes the `Error` from the response if one exists, replacing it with `None`.
+    pub fn take_error(&mut self) -> Option<Arc<Error>> {
```
Fishrock123, line 533: Alternatively, this could `Rc::try_unwrap()`, but I'm not really too sure about that.

Fishrock123:
Backed out 752e7e4 in favor of `Clone` with `error: None` + docs.

Fishrock123:
I've added `set_error()` here due to the discussion resulting in https://discordapp.com/channels/598880689856970762/649056551835009097/722181265435918469

This should cover all our bases. @yoshuawuyts Please take a look again when you are able to, I'd like to get this merged so that PRs on Tide (and soon Surf) can progress.

PR #178, opened 2020-06-07
Clippy: fix errors and enable ci

ririsoft:
Hello,

I am not sure about the project adherence to clippy. Reading #92 i understand it is accepted as a best practice with some exceptions such `new_without_default`.

I have thus fixed minor and straightforward clippy errors and warnings with a few exceptions described in the commit comment.

I have also added a github CI check to help keep the project clippy clean. Since clippy is consuming some precious minutes it can be moved to ubuntu only (as fmt checks) instead of all OSes if requested.

Cheers !


jbr:
Thanks!

PR #179, opened 2020-06-09
Mime: add From<&str>

Fishrock123:
Refs: https://github.com/http-rs/tide/issues/575

### DIFF
--- a/src/mime/mod.rs
+++ b/src/mime/mod.rs
@@ -139,6 +139,12 @@ impl FromStr for Mime {
     }
 }
 
+impl<'a> From<&'a str> for Mime {
+    fn from(value: &'a str) -> Self {
+        Self::from_str(value).unwrap()
+    }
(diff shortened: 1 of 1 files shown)


src/mime/mod.rs
```diff
@@ -139,6 +139,12 @@ impl FromStr for Mime {
     }
 }
 
+impl<'a> From<&'a str> for Mime {
+    fn from(value: &'a str) -> Self {
+        Self::from_str(value).unwrap()
```
yoshuawuyts, line 144: Could we use `expect` here instead? Probably copy over the message from `FromStr`'s error.

Fishrock123:
@yoshuawuyts I don't know how this can be improved aside from making it `TryFrom`. The parse error will always contain the amount of detail the user would actual need, and `.expect()` would obscure that?

PR #182, opened 2020-06-18
use serde_qs for querystrings

jbr:
This resolves a regression in tide: http-rs/tide#605

although serde_urlencoded is sufficient for key=value queries, it does not support the full range of queries that serde_qs does


jbr:
@yoshuawuyts I think we do need both unfortunately. serde_qs doesn't do `x-www-form-urlencoded`, which is needed for `Body::from_form`:

https://github.com/http-rs/http-types/blob/master/src/body.rs#L307

PR #188, opened 2020-06-25
mime: add as_str() to ParamName and ParamValue

Fishrock123:
Previously, there was no way to actually get the string values.

~~This (or something like it) is blocking using `http_types::mime` in Surf.~~

Noticed from Surf: https://github.com/http-rs/surf/blob/be10806bb9731c7a1d0f3d36296a1cc61267b775/src/response.rs#L164


Fishrock123:
The issue I ran into was ultimately solved by `.map(|v| v.to_string()).as_deref()`, but I feel like this would be better? i.e. `map(|v| v.as_str())`?

PR #190, opened 2020-06-29
add Body::make_head and AsMut<Body> for Response

jbr:
When responding to a HEAD request, body content should not be sent. Currently there is no way to retain the content length, but remove the actual content.

This PR:
* provides an affordance to strip out the content of the body but retain the content length.
* provides an `AsMut<Body>` for Response, which makes the use of make_head more convenient

I'm not happy with the name `make_head`, so please suggest improvements

refs: http-rs/tide#623

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -405,6 +405,11 @@ impl Body {
         self.length.map(|length| length == 0)
     }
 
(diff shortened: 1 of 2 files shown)


jbr:
Closing because async-h1 hangs indefinitely, so this needs to be addressed there

PR #197, opened 2020-07-08
expose Mime::from_extension

jbr:
this will be useful for an askama integration, and likely other places as well


Fishrock123:
We probably will want something along the lines of this at some point: https://www.npmjs.com/package/mime-db

PR #198, opened 2020-07-08
add Body::from_reader_with_mime

jbr:
This allows external integrations to build a Body that includes a mime type, which is otherwise impossible. An alternative would be exposing a `Body::set_mime`


Fishrock123:
I think we should do `Body::set_mime`?

yoshuawuyts:
`Body::{mime,set_mime}` feels better to me; if we feel this isn't enough we can always opt to add a shorthand constructor later as well.

PR #210, opened 2020-07-31
Cache control

yoshuawuyts:
Started work on Cache-Control headers, as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching. I feel like caching is something that could be really interesting to get right for Tide, and probably getting the typed headers in place is the first step to experiment with building more interesting abstractions.


src/cache/cache_control/cache_control.rs
```diff
@@ -0,0 +1,212 @@
...
+/// assert_eq!(entries.next().unwrap(), &CacheDirective::NoStore);
+/// #
+/// # Ok(()) }
+/// ```
+pub struct CacheControl {
+    entries: Vec<CacheDirective>,
+}
+
+impl CacheControl {
+    /// Create a new instance of `ServerTiming`.
```
Fishrock123, line 37:
```suggestion
    /// Create a new instance of `CacheControl`.
```

src/cache/cache_control/cache_control.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub fn new() -> Self {
+        Self { entries: vec![] }
+    }
+
+    /// Create a new instance from headers.
+    pub fn from_headers(headers: impl AsRef<Headers>) -> crate::Result<Option<Self>> {
```
Fishrock123, line 43: I thought `from_headers()` was supposed to just be `Option<Self>`?

And this doesn't actually seem to make use of `Result`?

(Although I think that in many cases `Result<Option<Header>>` will probably make sense but is so far inconsistent...)

src/cache/cache_control/cache_control.rs
```diff
@@ -0,0 +1,212 @@
...
+            inner: self.entries.into_iter(),
+        }
+    }
+}
+
+impl<'a> IntoIterator for &'a CacheControl {
+    type Item = &'a CacheDirective;
+    type IntoIter = Iter<'a>;
+
+    // #[inline]serv
```
Fishrock123, line 123:
```suggestion
    // #[inline]
```

src/cache/cache_control/cache_control.rs
yoshuawuyts: `from_headers` needs to be able to both detect whether a header was found, and whether parsing the header was successful. These are different code paths, and probably should be handled separately.

However in #99 an argument is made to use `Option<Result<Self>>` instead, which may actually be better. We should probably adjust all existing `from_header` methods to this shape, and update this one too (the only existing ones are currently behind an unstable flag, so we're okay).

src/cache/cache_control/cache_directive.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// The client will accept a stale response if retrieving a fresh one fails.
+    StaleIfError(Duration),
+    /// Indicates the client will accept a stale response, while asynchronously
+    /// checking in the background for a fresh one.
+    StaleWhileRevalidate(Duration),
+}
+
+impl CacheDirective {
+    /// Check whether this directive is valid in an HTTP request.
+    pub fn is_req(&self) -> bool {
```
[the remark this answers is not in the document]
yoshuawuyts:
```suggestion
    pub fn valid_in_req(&self) -> bool {
```

src/cache/cache_control/cache_directive.rs
```diff
@@ -0,0 +1,153 @@
...
+        use CacheDirective::*;
+        matches!(self,
+            MaxAge(_) | MaxStale(_) | MinFresh(_) | NoCache | NoStore | NoTransform
+            | OnlyIfCached)
+    }
+
+    /// Check whether this directive is valid in an HTTP response.
+    pub fn is_res(&self) -> bool {
```
yoshuawuyts, line 56:
```suggestion
    pub fn valid_in_res(&self) -> bool {
```

src/cache/cache_control/cache_control.rs
Fishrock123: On further thought I think that `Result<Option<T>>` should probably always be preferred, because it allows a seamless transition to `-> Option<T> throws Result` if that ever becomes a thing.

yoshuawuyts: The right signature would be `-> Option<T> throws Error`, but even then it'd be possible to write `-> crate::Result<T> throws NoneError`. I don't think the possibility of `throws` is something we should consider as a constraint for any design now. It's unclear if we'll ever have it, and even then the semantics are still undefined.

yoshuawuyts:
Going to merge this + the other PRs, and then do a pass in a separate PR to convert `Result<Option<Self>>` to `Option<Result<Self>>`.

PR #211, opened 2020-08-02
Typed ETags

yoshuawuyts:
Implements typed ETags. Builds on #210. Thanks!

Ref #99 

## Screenshots

Screenshot_2020-08-03 http_types cache - Rust


yoshuawuyts:
Note: Etags are also used for idempotent requests, not just caching. We should place this under a new "conditionals" submodule, the same as in the MDN header overview page.

yoshuawuyts:
Merging but not yet publishing; we need this for the `If-Modified-Since` and `If-Unmodified-Since` headers.

PR #212, opened 2020-08-04
enable error backtraces

jbr:
This is the result of a yakshave to get backtraces on http-types errors, which were not previously functional as documented.
`#[cfg(backtrace)]` does nothing by itself. Anyhow provides this flag to itself using [a build script](https://github.com/dtolnay/anyhow/blob/master/build.rs), so this PR adds a nearly-identical build script to http-types.

It also correctly forwards Debug fmt to the inner error.

PR #213, opened 2020-08-05
Track type name on error before losing that info

jbr:
If this turns out to have meaningful performance implications, we could feature flag this or just do it conditionally when debug_assertions are on.

This combines well with #212 to allow for this sort of development-mode error handler in tide:
image

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,6 +16,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 pub struct Error {
     error: anyhow::Error,
     status: crate::StatusCode,
+    type_name: Option<String>,
 }
 
 impl Error {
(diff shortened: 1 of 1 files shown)


jbr:
(I'm thinking the dev-mode-error-handler is something I probably could/should extract, once it's polished up. Not sure if it should live in tide or as an external crate, since it's somewhat opinionated, but assumedly opt-in)

src/error.rs
```diff
@@ -23,10 +24,11 @@ impl Error {
     /// The error type must be threadsafe and 'static, so that the Error will be
     /// as well. If the error type does not provide a backtrace, a backtrace will
     /// be created here to ensure that a backtrace exists.
-    pub fn new(status: StatusCode, error: impl Into<anyhow::Error>) -> Self {
+    pub fn new<E: Into<anyhow::Error>>(status: StatusCode, error: E) -> Self {
         Self {
             status,
             error: error.into(),
+            type_name: Some(std::any::type_name::<E>().to_string()),
```
Fishrock123, line 31: fun generic hehe

src/error.rs
```diff
@@ -15,6 +15,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 pub struct Error {
     error: anyhow::Error,
     status: crate::StatusCode,
+    type_name: Option<String>,
```
yoshuawuyts, line 18:
```suggestion
    type_name: Option<&'static str>,
```

src/error.rs
yoshuawuyts, line 31:
```suggestion
            type_name: Some(std::any::type_name::<E>()),
```

Fishrock123:
This looks merge-able?

PR #215, opened 2020-08-06
Eyre

jbr:
this is on top of the changes in #212 and not mutually exclusive with that or #213.

Tests are failing because somehow the Once is getting poisoned in tests. The Once is necessary because there's no obvious "app start" hook in http-types. One possibility is to just provide the backtrace method and require tide or some consumer to register stable-eyre. Or, ideally, we could just enable it with a feature flag on eyre, but eyre wants to be configured at runtime

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,6 @@ async-std = { version = "1.6.0", features = ["unstable"] }
 
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
-
-anyhow = "1.0.26"
 cookie = { version = "0.14.0", features = ["percent-encode"] }
 infer = "0.1.2"
 pin-project-lite = "0.1.0"
@@ -41,6 +39,8 @@ serde = { version = "1.0.106", features = ["derive"] }
 serde_urlencoded = "0.6.1"
 rand = "0.7.3"
 serde_qs = "0.6.0"
+backtrace = "0.3.50"
+stable-eyre = "0.2.1"
 
 [dev-dependencies]
 http = "0.2.0"
--- a/build.rs
+++ b/build.rs
@@ -3,9 +3,18 @@ use std::fs;
 use std::path::Path;
 use std::process::{Command, ExitStatus, Stdio};
 
-// This code exercises the surface area that we expect of the std Backtrace
-// type. If the current toolchain is able to compile it, we go ahead and use
-// backtrace in anyhow.
+// This build script is copied from
+// [anyhow](https://github.com/dtolnay/anyhow/blob/master/build.rs),
+// and is a type of feature detection to determine if the current rust
+// toolchain has backtraces available.
+//
+// It exercises the surface area that we expect of the std Backtrace
+// type. If the current toolchain is able to compile it, we enable a
+// backtrace compiler configuration flag in http-types. We then
+// conditionally require the compiler feature in src/lib.rs with
+// `#![cfg_attr(backtrace, feature(backtrace))]`
+// and gate our backtrace code behind `#[cfg(backtrace)]`
+
 const PROBE: &str = r#"
     #![feature(backtrace)]
     #![allow(dead_code)]
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,60 @@
+use std::env;
+use std::fs;
+use std::path::Path;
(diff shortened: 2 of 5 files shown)


yaahc:
ooh, as a library you'll probably want to avoid hard coding a dependency on one hook or another. I think this means I really should add back the object provider API in Eyre so libraries can do hook agnostic context lookup. Let me make some PRs to Eyre, stable-eyre, and color-eyre. then you'll need to use a diff fn to extract the backtrace but it will work for both stable-eyre and color-eyre.

Fishrock123:
What if we just use `eyre` directly as the error/result type and let e.g. Tide or the end application author install the hook?

That still allows the benefits and makes more structural sense. It also means that downstream could choose which reporter (such as color-eyre).

Additionally, perhaps we could make this a build option (feature)? We could then play around with it - I'm willing to try it out in my employer's application. I'm probably going to make my own branch to check things anyways, so I don't mind picking this up.

yaahc:
> What if we just use `eyre` directly as the error/result type and let e.g. Tide or the end application author install the hook?
> 
> That still allows the benefits and makes more structural sense. It also means that downstream could choose which reporter (such as color-eyre).
> 
> Additionally, perhaps we could make this a build option (feature)? We could then play around with it - I'm willing to try it out in my employer's application. I'm probably going to make my own branch to check things anyways, so I don't mind picking this up.

If you don't mind carrying the StatusCode along side the Report instead of inside, which this PR seems to imply already, then that's definitely the best option. I'm still a little sad about it. I wrote two PRs of different designs I was considering for how we could allow composable error reporting, or even extra handlers that you could push into reports from libraries, but neither of them really seemed like the right solution so they've kinda stalled.

- https://github.com/yaahc/eyre/pull/37
- https://github.com/yaahc/eyre/pull/38

My hope is that we'd eventually be able to consistently store the status code inside of the reporter, and retrieve it, and have little risk of users accidentally discarding them. Not there yet :/

Fishrock123:
@yaahc I noted in https://github.com/http-rs/http-types/pull/263#issuecomment-717487784 (my rebase of this PR) that changing the underlying type and then having a global hook doesn't really solve anything for Tide in particular, which is where this issue is most focused for me.

In Tide all runtime errors stay within Tide - any that end up at the response are converted into a `500 Internal Server Error`. To deal with runtime errors a logger middleware must be installed (one comes installed by default but can be disabled or replaced). It would have to handle the interaction with eyre reporting. Maybe you have some suggestions on how to best do that?

yaahc:
> In Tide all runtime errors stay within Tide - any that end up at the response are converted into a `500 Internal Server Error`. To deal with runtime errors a logger middleware must be installed (one comes installed by default but can be disabled or replaced). It would have to handle the interaction with eyre reporting. Maybe you have some suggestions on how to best do that?

Probably, but I'll need more details:

> to deal with runtime errors ...

Deal with how, just by logging or discarding them?

> It would have to handle the interaction with eyre reporting

might also need clarification here, depending on the answer to the previous question.

Fishrock123:
> Deal with how, just by logging or discarding them?

Tide turns all returns of `Result::Err(http_types::Error)` in middleware and endpoints immediately into a `tide::Response` with the `http_types::Error` attached. ([`http_types::Error](https://github.com/http-rs/http-types/blob/a6b2a056f7374eab0b5fdd5991ba09e0fc131a37/src/error.rs#L16-L20) is a wrapper around anyhow.)

A bare tide server does nothing with this information (on purpose, to avoid leaking internal error messages) other than use it's status code. The default log middleware, or a user log middleware, error body transform middleware, or some other mechanism, must grab an optional error off the request to check if there was one, and then can decide what to do with it (use something from it, downcast it, etc).

I think the default `LogMiddleware` with `error_eyre` enabled would probably be where something should happen. This is where we'd want to format it in some eyre custom way.

yaahc:
> I think the default `LogMiddleware` with `error_eyre` enabled would probably be where something should happen. This is where we'd want to format it in some eyre custom way.

Okay, I think I'm following so far. And what extra features did you need to get from the wrapped `eyre::Reports` that aren't currently covered by `anyhow::Report`? Is it that you need those errors to capture more context than anyhow can or is it that you want to customize the format used when logging errors with the default `LogMiddleware` or is it something else entirely?

PR #218, opened 2020-08-07
Remove Client and Server traits

yoshuawuyts:
Removes the experimental Client and Server traits. We won't be using Server for Tide, and the chance is slim we'll use Client for Surf. So it may be best to remove them from the project. Thanks!


jbr:
@Fishrock123 pointed out that this is not semver-major because it was unstable, so merging. My mistake!

PR #219, opened 2020-08-07
Cache age header

yoshuawuyts:
Adds the `cache::Age` header. [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Age). Ref #99 

## Screenshots

Screenshot_2020-08-07 http_types cache Age - Rust

### DIFF
--- a/src/cache/age.rs
+++ b/src/cache/age.rs
@@ -1,0 +1,117 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, AGE};
+use crate::Status;
+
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
One method this is missing is `duration` which provides access to the inner `Duration`. `fn duration(&self) -> Duration`

yoshuawuyts:
Merging but not yet publishing; we need the follow-up PR for the `If-Modified-Since` and `If-Unmodified-Since` headers.

PR #222, opened 2020-08-08
Add `conditional::{IfModifiedSince, IfUnmodifiedSince, LastModified}`

yoshuawuyts:
This adds two new conditional headers. Relies on #220. Ref #99. Thanks!


yoshuawuyts:
Realized `Last-Modified` is similar as well, so added that too!

src/conditional/if_unmodified_since.rs
```diff
@@ -0,0 +1,125 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, IF_MODIFIED_SINCE};
```
Fishrock123, line 1:
```suggestion
use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, IF_UNMODIFIED_SINCE};
```

(And other uses in this file)

Fishrock123 (changes requested):

PR #225, opened 2020-08-08
Add `conditional::Vary`

yoshuawuyts:
Implements the Vary header. Ref #99. Depends on #220, #224. This is actually the final conditional header, heh.

Vary takes a `HeaderName` as an input, so made it so it can take both header names and string literals for convenience. I think this should work reasonably well. Thanks!


yoshuawuyts:
The `VaryDirective` struct isn't needed; but that matches the rest of the directives and we can fix that in a follow-up PR.

PR #230, opened 2020-08-20
Add `content::{ContentEncoding,EncodingDirective}`

yoshuawuyts:
Ref #99; [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#Directives); adds structured `Content-Encoding` headers under the `content` submodule. Thanks!

## Further work

This needs a follow-up PR to specify the `Accept-Encoding` header which uses weighted encoding directives. This may require adjustments to the `EncodingDirective` (or a separate structure alltogether) so we should hold off on publishing this patch until we have both working.

## Screenshots

Screenshot_2020-08-20 http_types content - Rust

### DIFF
--- a/src/content/content_encoding.rs
+++ b/src/content/content_encoding.rs
@@ -1,0 +1,218 @@
+//! Specify the compression algorithm.
+
+use crate::content::EncodingDirective;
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
Superseded by #232.

PR #232, opened 2020-08-23
Add `content::AcceptEncoding`.

yoshuawuyts:
Supersedes #230, ref #99. Adds support for `Content-Encoding`, `Accept-Encoding` and content negotiation to select the right encoding.

The approach in this patch can likely be adapted to work for `TE`, `Accept-Language` and `Accept-Charset` as well.

## Screenshot

Screenshot_2020-08-23 http_types content AcceptEncoding - Rust


src/content/accept_encoding.rs
```diff
@@ -0,0 +1,429 @@
...
+    pub fn sort(&mut self) {
+        sort_by_weight(&mut self.entries);
+    }
+
+    /// Determine the most suitable `Content-Type` encoding.
+    ///
+    /// # Errors
+    ///
+    /// If no suitable encoding is found, an error with the status of `406` will be returned.
+    pub fn negotiate(&mut self, available: &[Encoding]) -> crate::Result<ContentEncoding> {
```
Fishrock123, line 115: It would be much usable if this instead returned `Result<Encoding>`. `tide-compress` is blocked from using this dur to that.

src/content/content_encoding.rs
```diff
@@ -0,0 +1,147 @@
...
+/// # Ok(()) }
+/// ```
+pub struct ContentEncoding {
+    inner: Encoding,
+}
+
+impl ContentEncoding {
+    /// Create a new instance of `CacheControl`.
+    pub fn new(encoding: Encoding) -> Self {
+        Self { inner: encoding }
```
Fishrock123, line 40: Additionally there should probably be a way to get the `Encoding` out from here.

src/content/content_encoding.rs
```diff
@@ -0,0 +1,147 @@
...
+                    inner = Some(entry);
+                }
+            }
+        }
+
+        let inner = inner.expect("Headers instance with no entries found");
+        Ok(Some(Self { inner }))
+    }
+
+    /// Sets the `Server-Timing` header.
```
Fishrock123, line 66:
```suggestion
    /// Sets the `Content-Encoding` header.
```

src/content/accept_encoding.rs
```diff
@@ -0,0 +1,429 @@
...
+                return Ok(encoding.into());
+            }
+        }
+
+        let mut err = Error::new_adhoc("No suitable ContentEncoding found");
+        err.set_status(StatusCode::NotAcceptable);
+        Err(err)
+    }
+
+    /// Insert a `HeaderName` + `HeaderValue` pair into a `Headers` instance.
```
Fishrock123, line 138:
```suggestion
    /// Sets the `Accept-Encoding` header.
```

src/content/accept_encoding.rs
```diff
@@ -0,0 +1,429 @@
...
+/// use http_types::content::{AcceptEncoding, ContentEncoding, Encoding, EncodingProposal};
+/// use http_types::Response;
+///
+/// let mut accept = AcceptEncoding::new();
+/// accept.push(EncodingProposal::new(Encoding::Brotli, Some(0.8))?);
+/// accept.push(EncodingProposal::new(Encoding::Gzip, Some(0.4))?);
+/// accept.push(EncodingProposal::new(Encoding::Identity, None)?);
+///
+/// let mut res = Response::new(200);
+/// let encoding = accept.negotiate(&[Encoding::Gzip, Encoding::Brotli])?;
```
Fishrock123, line 32:
```suggestion
/// let encoding = accept.negotiate(&[Encoding::Brotli, Encoding::Gzip])?;
```

Fishrock123 (dismissed):

src/content/accept_encoding.rs
Fishrock123: Ah it turns out you can `PartialEq` between them. Hmm, probably fine.

src/content/content_encoding.rs
```diff
@@ -0,0 +1,147 @@
...
+            for part in value.as_str().trim().split(',') {
+                // Try and parse a directive from a str. If the directive is
+                // unkown we skip it.
+                if let Some(entry) = Encoding::from_str(part) {
+                    inner = Some(entry);
+                }
+            }
+        }
+
+        let inner = inner.expect("Headers instance with no entries found");
```
Fishrock123, line 62: This probably shouldn't panic?

src/content/encoding.rs
```diff
@@ -0,0 +1,58 @@
+use crate::headers::HeaderValue;
+use std::fmt::{self, Display};
+
+/// Available compression algorithms.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum Encoding {
```
Fishrock123, line 7: It may be useful for this to have a `PartialEq<&str>`

e.g. `assert_eq!(res[headers::CONTENT_ENCODING], Encoding::Deflate);`

src/content/content_encoding.rs
yoshuawuyts: This should indeed never panic, but still adding the message just in case we ever mess up that constraint. Wish there was a better way to express: "If we have a list, there is always at least one item. We never create an empty list, that's `None` instead".

But ah well; can't have it all.

PR #233, opened 2020-08-24
headers: etag: reject invalid characters

ririsoft:
rfc7232 specifies a valid set of ASCII characters that etag headers should match.

This PR is adding the missing character validation.
Some similar implementation can be [found in Go](https://golang.org/src/net/http/fs.go#L307) for instance.

### DIFF
--- a/src/conditional/etag.rs
+++ b/src/conditional/etag.rs
@@ -84,6 +84,16 @@ impl ETag {
             }
         };
 
+        if !s
+            .bytes()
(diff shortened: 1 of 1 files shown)


ririsoft:
I see that the clippy fix is also submitted via #234 . 
I will then wait for it to be merged and rebase this PR on it if you agree.

Fishrock123:
@ririsoft I have merged the clippy fixes.

Fishrock123 (approved):

PR #234, opened 2020-08-24
src: fix clippy warnings

Fishrock123:
This should unblock CI on main.

### DIFF
--- a/src/trace/server_timing/metric.rs
+++ b/src/trace/server_timing/metric.rs
@@ -80,6 +80,7 @@ mod test {
     use std::time::Duration;
 
     #[test]
+    #[allow(clippy::redundant_clone)]
     fn encode() -> crate::Result<()> {
         let name = String::from("Server");
         let dur = Duration::from_secs(1);
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: clippy
-        args: -- -D warnings
+        args: --tests --examples -- -D warnings
 
     - name: tests
       uses: actions-rs/cargo@v1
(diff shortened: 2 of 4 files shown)


src/trace/server_timing/metric.rs
```diff
@@ -80,6 +80,7 @@ mod test {
     use std::time::Duration;
 
     #[test]
+    #[allow(clippy::redundant_clone)]
```
jbr, line 83: why allow instead of resolve? is this more complex than removing the last .clone()?

Fishrock123: Makes it more consistent to read.

Fishrock123:
Fixed something that was in an even newer version of clippy.

src/trace/server_timing/metric.rs
ririsoft: This one was probably catched by your editor but not by the CI which does not clippy the tests.
Maybe you could consider using `clippy --test` instead of `clippy` in the the CI to catch those issues earlier ?

Fishrock123: Huh? I run the same command as CI runs.

ririsoft: Strange indeed, On my side I only get the redundant closure error when running `cargo clippy --tests`. Running rust stable 1.45.2 and nightly 1.47.0-nightly (f44c6e4e2 2020-08-24).

ririsoft: Forgot to say I am running Windows 10, I will try on Linux tonight

ririsoft: Investigating this further I see that `cargo clippy --all-targets` also activate the `--tests` flag. This is maybe the command you are running ? When running only `cargo clippy -- -D warnings` (used by CI) I do not get the errors from `#[cfg(test)]` guarded code.

Fishrock123:
```
nightly-x86_64-pc-windows-msvc (overridden by +toolchain on the command line)
rustc 1.47.0-nightly (5180f3da5 2020-08-23)
```

```
cargo +nightly clippy --tests --examples -- -D warnings
```

Huh, turns out that's what Tide runs but not this. Adding a commit...

Fishrock123:
Ok I am merging this to unblock CI

PR #238, opened 2020-09-07
impl Display for Version

jbr:


src/version.rs
```diff
@@ -18,9 +18,34 @@ pub enum Version {
...
 }
 
+impl std::fmt::Display for Version {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Version::Http0_9 => "HTTP/0.9",
+            Version::Http1_0 => "HTTP/1.0",
+            Version::Http1_1 => "HTTP/1.1",
+            Version::Http2_0 => "HTTP/2.0",
+            Version::Http3_0 => "HTTP/3.0",
```
yoshuawuyts, line 28: Http 2 and 3 don't use a `.0` suffix when written out. Just looked it up and h2 used to have this, but the spec was changed to not use this.

yoshuawuyts (changes requested):

Fishrock123 (approved):

PR #240, opened 2020-09-16
Add `contains` to `HeaderValues`.

LukeMathWalker:
A bit of syntax sugar to avoid having to create an iterator and loop through it to check for membership.


jbr (approved):
Seems useful to me 👍

PR #246, opened 2020-09-19
Make async-std optional

### LINKED ISSUE #244: Make async-std dependency optional
Hello! Would it be possible to make the `async-std` dependency optional and feature-gated? Currently, `async-std` is used for the following: `async_std::io` traits, `async_std::sync::channel`, and `async_std::fs`. If `futures-lite` and `async-channel` are used instead (NOTE: these are already dependencies for `async-std`, so no new dependencies would be added), a bunch of transitive dependencies would be removed without sacrificing any functionality.

`Cargo.lock` diff, with `default-features = false`, after making `async-std` optional: https://gist.github.com/akhilles/60d553d797f9e0e94056da3274285e31.
It's also a fairly straight-forward change: https://gist.github.com/akhilles/29a8de956cf7beb274a2f8d6959d8e4f

Feel free to close this if it doesn't align with this library's design goals.

akhilles:
closes #244

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ cookie-secure = ["cookie/secure"]
 # features: async_std
 async-std = { version = "1.6.0", features = ["unstable"] }
 futures-lite = "1.7.0"
+async-channel = "1.4.2"
 
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,8 +27,8 @@ fs = ["async-std"]
 [dependencies]
 # features: async_std
 async-std = { version = "1.6.0", optional = true }
-futures-lite = "1.7.0"
-async-channel = "1.4.2"
+futures-lite = "1.11.1"
+async-channel = "1.5.1"
 
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ cookie-secure = ["cookie/secure"]
 # Note(yoshuawuyts): used for async_std's `channel` only; use "core" once possible.
 # features: async_std
 async-std = { version = "1.6.0", features = ["unstable"] }
+futures-lite = "1.7.0"
 
 # features: hyperium/http
 http = { version = "0.2.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,17 +16,17 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["async_std", "cookie-secure"]
+default = ["fs", "cookie-secure"]
 docs = ["unstable"]
 unstable = []
 hyperium_http = ["http"]
-async_std = [] # "async-std" when it is not default
+async_std = ["fs"]
 cookie-secure = ["cookie/secure"]
+fs = ["async-std"]
 
 [dependencies]
-# Note(yoshuawuyts): used for async_std's `channel` only; use "core" once possible.
 # features: async_std
-async-std = { version = "1.6.0", features = ["unstable"] }
+async-std = { version = "1.6.0", optional = true }
 futures-lite = "1.7.0"
 async-channel = "1.4.2"
 
@@ -46,4 +46,4 @@ serde_qs = "0.7.0"
 
 [dev-dependencies]
 http = "0.2.0"
-async-std = { version = "1.6.0", features = ["unstable", "attributes"] }
+async-std = { version = "1.6.0", features = ["attributes"] }
--- a/src/body.rs
+++ b/src/body.rs
@@ -357,7 +357,7 @@ impl Body {
     /// res.set_body(Body::from_file("/path/to/file").await?);
     /// # Ok(()) }) }
     /// ```
-    #[cfg(all(feature = "async_std", not(target_os = "unknown")))]
+    #[cfg(all(feature = "fs", not(target_os = "unknown")))]
     pub async fn from_file<P>(path: P) -> io::Result<Self>
     where
         P: AsRef<std::path::Path>,
@@ -477,7 +477,7 @@ impl AsyncBufRead for Body {
 
 /// Look at first few bytes of a file to determine the mime type.
 /// This is used for various binary formats such as images and videos.
-#[cfg(all(feature = "async_std", not(target_os = "unknown")))]
+#[cfg(all(feature = "fs", not(target_os = "unknown")))]
 async fn peek_mime(file: &mut async_std::fs::File) -> io::Result<Option<Mime>> {
     // We need to read the first 300 bytes to correctly infer formats such as tar.
     let mut buf = [0_u8; 300];
@@ -491,7 +491,7 @@ async fn peek_mime(file: &mut async_std::fs::File) -> io::Result<Option<Mime>> {
 
 /// Look at the extension of a file to determine the mime type.
 /// This is useful for plain-text formats such as HTML and CSS.
-#[cfg(all(feature = "async_std", not(target_os = "unknown")))]
+#[cfg(all(feature = "fs", not(target_os = "unknown")))]
 fn guess_ext(path: &std::path::Path) -> Option<Mime> {
     let ext = path.extension().map(|p| p.to_str()).flatten();
     ext.and_then(Mime::from_extension)
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,5 +1,4 @@
-use async_std::io::prelude::*;
-use async_std::io::{self, Cursor};
+use futures_lite::*;
 use serde::{de::DeserializeOwned, Serialize};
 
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -16,17 +16,19 @@ features = ["docs"]
...
 [features]
-default = ["async_std", "cookie-secure"]
+default = ["fs", "cookie-secure"]
 docs = ["unstable"]
 unstable = []
 hyperium_http = ["http"]
-async_std = [] # "async-std" when it is not default
+async_std = ["fs"]
 cookie-secure = ["cookie/secure"]
+fs = ["async-std"]
```
akhilles, line 25: I can remove the "fs" feature and just rely on "async_std", but I generally prefer feature names that communicate the capabilities that are being enabled by that feature. IMO, it makes it easier for the user to decide whether to enable/disable it.

src/body.rs
```diff
@@ -257,7 +256,7 @@ impl Body {
     /// # Examples
     ///
     /// ```
-    /// # fn main() -> Result<(), http_types::Error> { async_std::task::block_on(async {
+    /// # fn main() -> Result<(), http_types::Error> { futures_lite::future::block_on(async {
```
yoshuawuyts, line 259: Can we keep `async_std` in the examples?

src/upgrade/connection.rs
```diff
@@ -1,4 +1,4 @@
-use async_std::io::{self, prelude::*};
+use futures_lite::*;
```
yoshuawuyts, line 1: I don't love the wildcard imports for non-prelude crates; can we import the items individually instead?

yoshuawuyts (commented):
Two small nits on coding style that I'd like to see applied to the entirety of the patch; but otherwise a :100: change!

Cargo.toml
yoshuawuyts: I don't feel too strongly either way; happy to have an extra "fs" feature if you think that's a good idea!

src/body.rs
```diff
@@ -160,7 +159,7 @@ impl Body {
...
```
Fishrock123, line 162: Instead of relying on `block_on` (which `futures_util` does not have), these should move to using `#[async_std::main]`
```suggestion
    /// # #[async_std::main]
    /// # async fn main() -> Result<(), http_types::Error> {
```

src/upgrade/sender.rs
```diff
@@ -1,28 +1,26 @@
...
 pub struct Sender {
-    sender: sync::Sender<Connection>,
+    sender: async_channel::Sender<Connection>,
 }
 
 impl Sender {
     /// Create a new instance of `Sender`.
     #[doc(hidden)]
-    pub fn new(sender: sync::Sender<Connection>) -> Self {
+    pub fn new(sender: async_channel::Sender<Connection>) -> Self {
```
Fishrock123, line 16: Is this a breaking change? This may be doc-hidden but if async-h1 depends on it there could be problems.

Fishrock123 (changes requested):

Cargo.toml
```diff
@@ -16,17 +16,19 @@ features = ["docs"]
...
 
 [dependencies]
-# Note(yoshuawuyts): used for async_std's `channel` only; use "core" once possible.
 # features: async_std
-async-std = { version = "1.6.0", features = ["unstable"] }
+async-std = { version = "1.6.0", optional = true }
+futures-lite = "1.7.0"
```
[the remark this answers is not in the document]
yoshuawuyts: > Could this use futures-util instead?

Has it? `futures-util` is a fairly heavy dependency that has a lot of overlap with what `async-std` provides. I would prefer using `futures-lite` over `futures-util` in most cases.

src/upgrade/sender.rs
yoshuawuyts: I don't believe `async-h1` depends on it ([search results](https://github.com/http-rs/async-h1/search?q=Sender%3A%3Anew&unscoped_q=Sender%3A%3Anew)); if I recall correctly this is only exposed so that we can use it in integration tests. You're right though that if we have another crate depending on this it would complicate things..

src/body.rs
akhilles: fixed

src/upgrade/connection.rs
akhilles: fixed

src/body.rs
akhilles: Replaced with `async_std::task::block_on` since that was already being used.

src/upgrade/sender.rs
akhilles: This does seem to be a breaking change. I don't why anyone would use `Sender::new` since there's no way to add that sender to a `Request`/`Response` type, but 🤷. It's also gated behind the "unstable" feature, and I couldn't find any dependent crates that enable this.

Maybe it's still a good idea to wait for the next breaking release?

akhilles: I would also suggest making this function private, since it has no real use in the public API.

Cargo.toml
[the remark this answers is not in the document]
Fishrock123: Hmm, ok, news to me. I'll look into that on the other repos then.

src/body.rs
Fishrock123: These should still be moved to `#[async_std::main]`, either in this PR or in another.

src/upgrade/sender.rs
Fishrock123: I believe only the upgrade sender stuff is behind `"unstable"`, were it all we could do that change with less worry.

akhilles: Only `upgrade::Sender::new` is a breaking change right? Nothing outside "unstable" should be affected.

akhilles: Just noticed `trailers::Sender::new` and `trailers::Receiver::new` are outside "unstable" and they're affected too. But I don't think anyone would use those functions outside of `http-types`; they're completely useless unless they're in a `Request`/`Response` type.

src/body.rs
akhilles: Using `#[async_std::main]` does increase compile times though. It's probably better to make that change in a separate PR.

src/upgrade/sender.rs
yoshuawuyts: both `trailers::Sender::new` and `upgrade::Sender::new` are non-public methods; neither of these affects our stability

src/body.rs
Fishrock123: Ah I suppose we may want to leave the hidden ones as is then...

jbr:
This looks good. Is it semver-major?

akhilles:
> This looks good. Is it semver-major?

I don't believe so.

akhilles:
`futures-lite` now has a prelude, updated the imports to use it.

src/upgrade/receiver.rs
```diff
@@ -8,20 +7,16 @@ use std::task::{Context, Poll};
 use crate::upgrade::Connection;
 
 /// The receiving half of a channel to send an upgraded connection.
-///
-/// Unlike `async_std::sync::channel` the `send` method on this type can only be
-/// called once, and cannot be cloned. That's because only a single instance of
-/// `Connection` should be created.
 #[must_use = "Futures do nothing unless polled or .awaited"]
```
jbr, line 10: Is this comment no longer relevant? Why was it removed?

src/upgrade/sender.rs
```diff
@@ -1,28 +1,26 @@
...
         Self { sender }
     }
 
     /// Send a `Trailer`.
     ///
     /// The channel will be consumed after having sent trailers.
     pub async fn send(self, trailers: Connection) {
-        self.sender.send(trailers).await
+        let _ = self.sender.send(trailers).await;
```
jbr, line 24: This line seems worth discussing. async-std channels cannot be closed as long as there's a sender, but async-channels can be, and return an Err variant if the channel is closed. Probably the best fix for this is to return that Result, making this semver-major. As written, this silently swallows the possible SendError.

jbr (dismissed):
I'm excited to get this merged, but have concerns about the change in async channel semantics regarding send errors and closed channels

src/upgrade/receiver.rs
akhilles: This is the receiver, the comment was meant for the sender I assume.

src/upgrade/sender.rs
akhilles: Functionally, this is the same as existing behavior. It's a oneshot channel where the sender doesn't care if the receiver is dropped. It might be a good idea to return a `Result` here if the receiver has been dropped, but I see it as a separate issue.

akhilles: Since this `Sender`/`Receiver` code is duplicated a couple times, I'd recommend making a generic `oneshot` submodule that's shared by `trailers`, `upgrade`, and whatever else relies on this pattern.

akhilles: I was planning that as a follow-up, but I wanted to keep this PR non-breaking.

PR #247, opened 2020-09-19
Remove directives

### LINKED ISSUE #241: TimingAllowOrigin should not operate on TimingOrigin
This is unreleased and blocking the next release since we should change it. But currently the `TimingAllowOrigin` looks like this:

```rust
    #[test]
    fn smoke() -> crate::Result<()> {
        let mut origins = TimingAllowOrigin::new();
        origins.push(TimingOrigin::Wildcard);

        let mut headers = Headers::new();
        origins.apply(&mut headers);

        let origins = TimingAllowOrigin::from_headers(headers)?.unwrap();
        let origin = origins.iter().next().unwrap();
        assert_eq!(origin, &TimingOrigin::Wildcard);
        Ok(())
    }
```

Accept-Encoding is coming in #232, and the instead of wrapping a value in an enum we add a method exposing whether a `wildcard` argument was passed:

```rust
    #[test]
    fn wildcard() -> crate::Result<()> {
        let mut accept = AcceptEncoding::new();
        accept.set_wildcard(true);

        let mut headers = Response::new(200);
        accept.apply(&mut headers);

        let accept = AcceptEncoding::from_headers(headers)?.unwrap();
        assert!(accept.wildcard());
        Ok(())
    }
```

This means that `AcceptEncoding`'s `EncodingProposal` type doesn't need a wrapper or extra field on the type to signal it accepts a wildcar: that's just a state of the header, and something we take in account during content resolution.

We should make the same change for `TimingAllowOrigin`: it should take and iterate over `Url` instances rather than the `TimingOrigin` wrapper type. And expose two new methods:

- `wildcard`
- `set_wildcard`

We could even add some validation method to check if a given url is allowed; but given we want to make a release sooner rather than later this is the minimal change we need to do.

yoshuawuyts:
Removes all intermediate `Directive*` types in favor of a `wildcard` and `set_wildcard` method on the containing type. This closes #241, but also addresses the same design flaw in the `conditional` series of PRs we just merged. Thanks!


yoshuawuyts:
Oh also worth noting: this patch only changes unreleased APIs so it isn't a breaking change.

Fishrock123 (approved):
I kinda thought the old api was neat, but I don't feel strongly about it.

yoshuawuyts:
> I kinda thought the old api was neat, but I don't feel strongly about it.

The old API made it tricky to implement comparison methods; e.g. "does this etag match" was quite awkward and required keeping a temporary var to see if globs are accepted. Instead by tracking globs at compile-time this logic is simpler, and we can remove the intermediate structs.

PR #249, opened 2020-09-27
Allow `Error::set_status` to take `TryInto<StatusCode>`

yoshuawuyts:
This is a minor annoyance when working with status codes. This brings it in line with the rest of our methods. Thanks!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -69,8 +69,14 @@ impl Error {
     }
 
     /// Set the status code associated with this error.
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Fix for clippy in https://github.com/http-rs/http-types/pull/250. Waiting for PR to pass so we can rebase.

yoshuawuyts:
Oki, merging. The new release contains many similar APIs and this was mostly left behind on accident it seems.

Issue #227, opened 2020-08-11
macros should take a status code as their second argument


yoshuawuyts:
Using `ensure!` and pals currently always yield a 500 internal server errror. We should make it so that an `Into<StatusCode>` must always be provided so these actually function as shorthands:

```rust
ensure!(tweet.len() < 140, 400, "Tweet must be under 140 characters");
```

jbr:
It's not entirely clear what the second arg is in this context, but since we're in a macro context, what if it took named arguments like an optional `status: 400`, as in

`ensure!(tweet.len() < 140, "Tweet must be under 140 characters", status: 400);`

I think that should be possible?

yoshuawuyts:
I found [jshttp/http-errors]() recently, which uses a very similar API to the one I proposed:

```js
var err = createError(404, 'This video does not exist!')
```

I think making the status code always required is not a bad idea, especially since they're convenient to author now that we can convert them from any number. So the equivalent JS code in http-types would look like:

```rust
let err = format_err!(404, "This video does not exist");
```

yoshuawuyts:
Prior art for this exists in https://github.com/jshttp/http-assert and https://github.com/jshttp/http-errors.

PR #252, opened 2020-10-01
Add `auth::{Authorization, AuthenticationScheme, BasicAuth, WwwAuthenticate, ProxyAuthorization, ProxyAuthenticate}`

yoshuawuyts:
Ref #99 

Adds HTTP basic auth and related structs for authentication.

BasicAuth is a specialization of `Authorization`; because schemes like oauth are rather elaborate, we've chosen to put them into separate structs.

Thanks!

## Screenshots

Screenshot_2020-10-01 http_types auth - Rust

### DIFF
--- a/src/auth/basic_auth.rs
+++ b/src/auth/basic_auth.rs
@@ -19,7 +19,7 @@ use crate::Status;
 ///
 /// let username = "nori";
 /// let password = "secret_fish!!";
-/// let authz = BasicAuth::new(username, Some(password));
+/// let authz = BasicAuth::new(username, password);
 ///
 /// let mut res = Response::new(200);
 /// authz.apply(&mut res);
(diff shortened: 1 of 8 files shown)


yoshuawuyts:
I've added support for `WWW-Authenticate`, but need to change it up slightly before merging. Right now we support at most 1 authentication scheme, but the spec says this should be up to n. Also since `Proxy-Authenticate` parses *exactly* the same way as `WWW-Authenticate` we can copy-paste the impl once it works. That should provide the majority of common authentication things.

yoshuawuyts:
Okay; all implemented!

Fishrock123:
`BasicAuth` seems to work fine, as I am using it here: https://github.com/squamishaccess/squamishaccess-signup-function-rs/blob/6c445916760a4db21da80c732fbe8e9c704c1bd1/src/main.rs#L168-L172

However I did run into a UX issue with typed headers and surf: https://github.com/http-rs/surf/issues/251

src/auth/basic_auth.rs
```diff
@@ -0,0 +1,141 @@
...
+        let bytes = base64::decode(auth.credentials()).status(400)?;
+        let credentials = String::from_utf8(bytes).status(400)?;
+
+        let mut iter = credentials.splitn(2, ':');
+        let username = iter.next();
+        let password = iter.next();
+
+        let (username, password) = match (username, password) {
+            (Some(username), Some(password)) => (username.to_string(), password.to_string()),
+            (Some(_), None) => bail!(400, "Expected basic auth to a password"),
```
Fishrock123, line 76:
```suggestion
            (Some(_), None) => bail!(400, "Expected basic auth to contain a password"),
```

src/macros.rs
```diff
@@ -84,3 +84,99 @@ macro_rules! format_err {
         $crate::private::new_adhoc(format!($fmt, $($arg)*))
     };
 }
+
+/// Return early with an error and a status code.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! bail2 {
```
Fishrock123, line 91: Maybe we should consider calling these new versions `{}_status`?

Fishrock123 (approved):
For some reason I had the impression this covered more than basic auth. Looks good, although I am not familiar with `WWW-Authenticate`.

Basic auth certainly works when I make client requests to mailchimp.

One question I have is how to make Authorization and Proxy-Authorization work seamlessly.

src/macros.rs
yoshuawuyts: Sure thing! Though they're only internal for now which means the naming doesn't matter too much -- for the next major we should probably make these the default as per https://github.com/http-rs/http-types/issues/227

PR #253, opened 2020-10-06
Add `content::ContentLength`

yoshuawuyts:
ref #99 adds the typed `Content-Length` header. Thanks!

### DIFF
--- a/src/content/content_length.rs
+++ b/src/content/content_length.rs
@@ -30,6 +30,7 @@ pub struct ContentLength {
     length: usize,
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl ContentLength {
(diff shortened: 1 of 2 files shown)


src/content/content_length.rs
```diff
@@ -0,0 +1,108 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_LENGTH};
+use crate::Status;
+
+/// Credentials to authenticate a user agent with a server.
```
Fishrock123, line 4: this is from the Auth pr

Fishrock123 (commented):
One little docs thing.

Could almost have a macro for some of these, hah.

Fishrock123 (approved):

PR #256, opened 2020-10-08
Add `content::ContentLocation` header

pepoviola:
Hi All, this is an attempt to make my small contribution to this awesome project 🙌 . This PR is inspired by https://github.com/http-rs/http-types/pull/253 ( I try to following the same approach ) and also ref to one of the headers listed on #99.

I have some question related to the field for the `ContentLocation` struct. Is ok to be a `String` and parse with the `Url` crate or is better to use another type ( like to be directly an `url` and just format in the `value` ) ? 

I'm new to rust and my code could be partial ( or totally 😄  ) wrong, so any feedback is welcome.

### DIFF
--- a/src/content/content_location.rs
+++ b/src/content/content_location.rs
@@ -1,0 +1,107 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_LOCATION};
+use crate::{Status, Url};
+
+///
+/// # Specifications
+///
+/// - [RFC 7230, section 3.3.2: Content-Length](https://tools.ietf.org/html/rfc7230#section-3.3.2)
+///
+/// # Examples
+///
+/// ```
+/// # fn main() -> http_types::Result<()> {
+/// #
+/// use http_types::Response;
+/// use http_types::content::{ContentLocation};
+///
+/// let content_location = ContentLocation::new("https://example.net/".to_string());
+///
+/// let mut res = Response::new(200);
+/// content_location.apply(&mut res);
+///
+/// let content_location = ContentLocation::from_headers(res)?.unwrap();
+/// assert_eq!(content_location.location(), "https://example.net/");
+/// #
+/// # Ok(()) }
+/// ```
+#[derive(Debug)]
+pub struct ContentLocation {
+    url: String,
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl ContentLocation {
+    /// Create a new instance of `Content-Location` header.
+    pub fn new(url: String) -> Self {
+        Self { url }
+    }
+
+    /// Create a new instance from headers.
+    pub fn from_headers(headers: impl AsRef<Headers>) -> crate::Result<Option<Self>> {
+        let headers = match headers.as_ref().get(CONTENT_LOCATION) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let value = headers.iter().last().unwrap();
+        let url = Url::parse(value.as_str().trim()).status(400)?;
+        Ok(Some(Self { url : url.into_string() }))
+    }
+
+    /// Sets the header.
+    pub fn apply(&self, mut headers: impl AsMut<Headers>) {
+        headers.as_mut().insert(self.name(), self.value());
+    }
+
+    /// Get the `HeaderName`.
+    pub fn name(&self) -> HeaderName {
+        CONTENT_LOCATION
+    }
+
+    /// Get the `HeaderValue`.
+    pub fn value(&self) -> HeaderValue {
+        let output = format!("{}", self.url);
+
+        // SAFETY: the internal string is validated to be ASCII.
+        unsafe { HeaderValue::from_bytes_unchecked(output.into()) }
+    }
+
+    /// Get the url.
+    pub fn location(&self) -> String {
+        String::from(&self.url)
+    }
+
+    /// Set the url.
+    pub fn set_location(&mut self, location: &str) {
+        self.url = location.to_string();
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::headers::Headers;
+
+    #[test]
+    fn smoke() -> crate::Result<()> {
+        let content_location = ContentLocation::new("https://example.net/test".to_string());
+
+        let mut headers = Headers::new();
+        content_location.apply(&mut headers);
+
+        let content_location = ContentLocation::from_headers(headers)?.unwrap();
+        assert_eq!(content_location.location(), "https://example.net/test");
+        Ok(())
+    }
+
+    #[test]
+    fn bad_request_on_parse_error() -> crate::Result<()> {
+        let mut headers = Headers::new();
+        headers.insert(CONTENT_LOCATION, "<nori ate the tag. yum.>");
+        let err = ContentLocation::from_headers(headers).unwrap_err();
+        assert_eq!(err.status(), 400);
+        Ok(())
+    }
+}
--- a/src/content/content_location.rs
+++ b/src/content/content_location.rs
@@ -22,7 +22,8 @@ use std::convert::TryInto;
 /// let mut res = Response::new(200);
 /// content_location.apply(&mut res);
 ///
-/// let content_location = ContentLocation::from_headers(Url::parse("https://example.net/").unwrap(),res)?.unwrap();
+/// let url = Url::parse("https://example.net/")?;
+/// let content_location = ContentLocation::from_headers(url, res)?.unwrap();
 /// assert_eq!(content_location.location(), "https://example.net/");
 /// #
 /// # Ok(()) }
--- a/src/content/content_location.rs
+++ b/src/content/content_location.rs
@@ -52,7 +52,7 @@ impl ContentLocation {
         // If we successfully parsed the header then there's always at least one
         // entry. We want the last entry.
         let value = headers.iter().last().unwrap();
-        let base = base_url.try_into().expect("Could not convert into a valid url");
+        let base = base_url.try_into()?;
         let url = base.join(value.as_str().trim()).status(400)?;
         Ok(Some(Self { url }))
     }
--- a/src/content/content_location.rs
+++ b/src/content/content_location.rs
@@ -1,4 +1,5 @@
 use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_LOCATION};
+use crate::{Error, StatusCode};
 use crate::{Status, Url};
 
 use std::convert::TryInto;
@@ -53,14 +54,14 @@ impl ContentLocation {
         // If we successfully parsed the header then there's always at least one
         // entry. We want the last entry.
         let value = headers.iter().last().unwrap();
-        let base = base_url.try_into().ok();
-        match base {
-            Some(base) => {
-                let url = base.join(value.as_str().trim()).status(400)?;
-                Ok(Some(Self { url }))
-            }
-            None => Ok(None),
-        }
+        let base = base_url.try_into().map_err(|_| {
+            let mut err = Error::new_adhoc("Invalid base url provided");
+            err.set_status(StatusCode::BadRequest);
+            err
+        })?;
+
+        let url = base.join(value.as_str().trim()).status(400)?;
+        Ok(Some(Self { url }))
     }
 
     /// Sets the header.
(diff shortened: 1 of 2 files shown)


yoshuawuyts (commented):
Thanks so much for this PR! -- this is really good progress!

> I have some question related to the field for the ContentLocation struct. Is ok to be a String and parse with the Url crate or is better to use another type ( like to be directly an url and just format in the value ) ?

I think we should be operating on a `Url` and then converting that to a string internally. This does mean that `from_headers` may need to take the form of `from_headers(base_url: impl TryInto<Url>, headers: impl AsRef<Headers>)` so that a valid `Url` can always be constructed, because `Content-Location` may send either a full URL or just a pathname. That way end-users will always get access to a full URL, which removes the difference entirely.

yoshuawuyts:
Also ref #99

pepoviola:
Hi @yoshuawuyts, Thanks for the feedback! I make some of the required changes ( the small ones ) and also change the struct to use `Url` type internally. 

I'm not sure if this is ok, since an invalid base_url should return a `400` error right?
```
        let base = base_url.try_into().expect("Could not convert into a valid url");
        let url = base.join(value.as_str().trim()).status(400)?;
```

Again thanks for the feedback, and let me know if the approach isn't good.

src/content/content_location.rs
```diff
@@ -0,0 +1,114 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_LOCATION};
+use crate::{Status, Url};
+
+use std::convert::TryInto;
+
+/// Indicates an alternate location for the returned data.
+///
+/// # Specifications
+///
+/// - [RFC 7231, section 3.1.4.2: Content-Length](https://tools.ietf.org/html/rfc7231#section-3.1.4.2)
```
Fishrock123, line 10: should be the RFC for `Content-Location`

src/content/content_location.rs
```diff
@@ -0,0 +1,114 @@
...
+/// # Ok(()) }
+/// ```
+#[derive(Debug)]
+pub struct ContentLocation {
+    url: Url,
+}
+
+impl ContentLocation {
+    /// Create a new instance of `Content-Location` header.
+    pub fn new(url: String) -> Self {
```
Fishrock123, line 37: I think it would be more ideal if this took a `Url` or `impl TryInto<Url>`.

src/content/content_location.rs
```diff
@@ -0,0 +1,114 @@
...
+   {
+        let headers = match headers.as_ref().get(CONTENT_LOCATION) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let value = headers.iter().last().unwrap();
+        let base = base_url.try_into().expect("Could not convert into a valid url");
```
Fishrock123, line 55:
```suggestion
        let base = base_url.try_into()?;
```
(Let's not crash a server on foreign input.)

src/content/content_location.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Get the url.
+    pub fn location(&self) -> String {
+        self.url.to_string()
+    }
+
+    /// Set the url.
+    pub fn set_location(&mut self, location: &str) {
+        self.url = Url::parse(location).unwrap();
```
Fishrock123, line 85: Same as `::new()`, probably more important here.

src/content/content_location.rs
pepoviola: Hi @Fishrock123, thanks for the feedback! yes, that sound good 👍 
Thx!

src/content/content_location.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl ContentLocation {
+    /// Create a new instance of `Content-Location` header.
+    pub fn new(url: Url) -> Self {
+        Self { url }
+    }
+
+    /// Create a new instance from headers.
+    pub fn from_headers<U>(base_url: U, headers: impl AsRef<Headers>) -> crate::Result<Option<Self>>
```
Fishrock123, line 42: Is this correct? (Mostly hoping Yosh with double check this.)

Fishrock123 (approved):
This looks good to me aside from one question.

pepoviola:
Thanks @Fishrock123!  I will open the pr for review 👍

src/content/content_location.rs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        let headers = match headers.as_ref().get(CONTENT_LOCATION) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let value = headers.iter().last().unwrap();
+        let base = base_url.try_into().unwrap();
```
Fishrock123, line 55: Oh this still doesn't seem quite right. I had meant this:
```suggestion
        let base = base_url.try_into()?;
```

src/content/content_location.rs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        let headers = match headers.as_ref().get(CONTENT_LOCATION) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let value = headers.iter().last().unwrap();
+        let base = base_url.try_into()?;
```
Fishrock123, line 55: (This probably requires some kind of `map_err()` as noted by the build failures.)

pepoviola: Hi @Fishrock123, I use `ok` here and a match later. I don't know if it's the best approach.
Thx!

src/content/content_location.rs
```diff
@@ -0,0 +1,123 @@
...
+/// #
+/// use http_types::{Response,Url};
+/// use http_types::content::{ContentLocation};
+///
+/// let content_location = ContentLocation::new(Url::parse("https://example.net/")?);
+///
+/// let mut res = Response::new(200);
+/// content_location.apply(&mut res);
+///
+/// let content_location = ContentLocation::from_headers(Url::parse("https://example.net/").unwrap(),res)?.unwrap();
```
yoshuawuyts, line 25:
```suggestion
/// let url = Url::parse("https://example.net/")?;
/// let content_location = ContentLocation::from_headers(url, res)?.unwrap();
```

src/content/content_location.rs
yoshuawuyts: @pepoviola I think what @Fishrock123 meant is that if you call `base_url.try_into()?` then by default a status code of 500 will be set on the error. The right solution would indeed be to call `map_err` and convert the error to a `400` status code. Returning `None` ignores the error, which is less good than having a 500 status code.

Instead the way to address this would be:

```rust
        // If we successfully parsed the header then there's always at least one
        // entry. We want the last entry.
        let value = headers.iter().last().unwrap();
        let base = base_url.try_into().map_err(|mut e| {
            e.set_status(400);
            e
        })?;
```

Hope that's helpful!

src/content/content_location.rs
yoshuawuyts: I think it is!

yoshuawuyts (commented):
a few more notes

src/content/content_location.rs
pepoviola: Hi @yoshuawuyts, thanks for the feedback!! Yes, I wasn't sure how to mapping the errors but now make a lot of sense.
I will change to your approach. Thanks for the help!

pepoviola: Hi @yoshuawuyts, sorry to bother you but I changed to the above approach and now I getting this errors related to the Error type and the `set_status` method.

```rust
error[E0599]: no method named `set_status` found for associated type `<U as std::convert::TryInto<url::Url>>::Error` in the current scope
  --> src/content/content_location.rs:58:15
   |
58 |             e.set_status(400);
   |               ^^^^^^^^^^ method not found in `<U as std::convert::TryInto<url::Url>>::Error`

error[E0277]: the trait bound `<U as std::convert::TryInto<url::Url>>::Error: std::error::Error` is not satisfied
  --> src/content/content_location.rs:60:11
   |
60 |         })?;
   |           ^ the trait `std::error::Error` is not implemented for `<U as std::convert::TryInto<url::Url>>::Error`
   |
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `anyhow::Error`
   = note: required because of the requirements on the impl of `std::convert::Into<anyhow::Error>` for `<U as std::convert::TryInto<url::Url>>::Error`
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `error::Error`
   = note: required by `std::convert::From::from`
help: consider further restricting the associated type
   |
46 |         U::Error: std::fmt::Debug, <U as std::convert::TryInto<url::Url>>::Error: std::error::Error
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: `<U as std::convert::TryInto<url::Url>>::Error` cannot be sent between threads safely
  --> src/content/content_location.rs:60:11
   |
60 |         })?;
   |           ^ `<U as std::convert::TryInto<url::Url>>::Error` cannot be sent between threads safely
   |
   = help: the trait `std::marker::Send` is not implemented for `<U as std::convert::TryInto<url::Url>>::Error`
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `anyhow::Error`
   = note: required because of the requirements on the impl of `std::convert::Into<anyhow::Error>` for `<U as std::convert::TryInto<url::Url>>::Error`
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `error::Error`
   = note: required by `std::convert::From::from`
help: consider further restricting the associated type
   |
46 |         U::Error: std::fmt::Debug, <U as std::convert::TryInto<url::Url>>::Error: std::marker::Send
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: `<U as std::convert::TryInto<url::Url>>::Error` cannot be shared between threads safely
  --> src/content/content_location.rs:60:11
   |
60 |         })?;
   |           ^ `<U as std::convert::TryInto<url::Url>>::Error` cannot be shared between threads safely
   |
   = help: the trait `std::marker::Sync` is not implemented for `<U as std::convert::TryInto<url::Url>>::Error`
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `anyhow::Error`
   = note: required because of the requirements on the impl of `std::convert::Into<anyhow::Error>` for `<U as std::convert::TryInto<url::Url>>::Error`
   = note: required because of the requirements on the impl of `std::convert::From<<U as std::convert::TryInto<url::Url>>::Error>` for `error::Error`
   = note: required by `std::convert::From::from`
help: consider further restricting the associated type
   |
46 |         U::Error: std::fmt::Debug, <U as std::convert::TryInto<url::Url>>::Error: std::marker::Sync
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 4 previous errors
```

I try adding the restrictions but don't seems like the best path, right?

Thanks!

src/content/content_location.rs
```diff
@@ -0,0 +1,125 @@
...
+        let headers = match headers.as_ref().get(CONTENT_LOCATION) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let value = headers.iter().last().unwrap();
+        let base = base_url.try_into().map_err(|_| {
+            let mut err = Error::new_adhoc("Invalid base url provided");
```
pepoviola, line 58: Hi @yoshuawuyts, I changed to use here `Error::new_adhoc` not sure if it's the best way to return `400`.
Thanks!

yoshuawuyts: @pepoviola you can use `crate::bail_status` to return a new error with a 400 status code

pepoviola: Hi @yoshuawuyts, thanks for the help! I changed to use `bail_status`.
Thx!

yoshuawuyts (approved):
:100: this looks great; thank you!

pepoviola:
awesome!, thanks for the guide and all the help!! 🙌

PR #258, opened 2020-10-08
error: allow backtrace Option to be Display on stable

Fishrock123:
This prevents the need for users to do backtrace feature-checking for code that wishes to use `Display` for a backtrace and also compile on stable.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -19,6 +19,17 @@ pub struct Error {
     type_name: Option<&'static str>,
 }
 
+#[allow(unreachable_pub)]
+#[derive(Debug)]
+#[doc(hidden)]
+pub struct BacktracePlaceholder;
+
+impl Display for BacktracePlaceholder {
+    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
+        unreachable!()
+    }
+}
+
 impl Error {
     /// Create a new error object from any error type.
     ///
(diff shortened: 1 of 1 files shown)


jbr:
Why wouldn't the user just test if the option is None?

jbr:
Ohhh… I see what's going on. What if we just used some already available type like Option<String> (which also implements Display)

Fishrock123:
I don't understand? `Option<Option<?>>`? `Option` only implements `Display` if `T` also does.
The idea here is to have the same traits as `Backtrace` so the type system doesn't complain on stable.

jbr:
@Fishrock123 sorry, forgot that comments don't auto escape `<` — typed `Option<String>`. the String part is what would change

Fishrock123:
I would rather it be a type that implements exactly the same traits as `Backtrace`, personally.

Fishrock123:
Ultimately, the correct answer here is that it would be `Option<!>`, using the [never](https://doc.rust-lang.org/std/primitive.never.html) type to denote that it will only ever be `None`. (And also because `never` is able to auto-implement all basic traits. Or something.)

jbr:
Yep, agreed that `!` would be ideal if stable

yoshuawuyts (approved):
Oh, this is a fun patch. Looking good!

PR #259, opened 2020-10-15
convert an http_types::Error into an anyhow::Error

jbr:
Related to https://github.com/http-rs/surf/issues/86

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -112,6 +112,12 @@ impl Error {
         None
     }
 
+    /// Returns the inner [`anyhow::Error`]
+    /// Note: This will lose status code information
+    pub fn into_inner(self) -> anyhow::Error {
+        self.error
+    }
+
     /// Attempt to downcast the error object to a concrete type.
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -163,7 +169,6 @@ impl<E: Into<anyhow::Error>> From<E> for Error {
         Self::new(StatusCode::InternalServerError, error)
     }
 }
-
```
Fishrock123, line 166: this seems accidental

Fishrock123 (approved):
Seems reasonable.

faulesocke:
The only thing required for interop with anyhow and similar crates is an implementation of `std::error::Error`. No dependency on anyhow or anything else is required.

Fishrock123:
@jbr Any reason you haven't merged this?

jbr:
Nope just keep forgetting to remove the newline you pointed out when I'm at a keyboard. Meeting as-is

PR #262, opened 2020-10-27
Add custom deserialize and serialize impls for Version, StatusCode and Method

### LINKED ISSUE #260: Derive Serializable and Deserializable for http_types::method::Method
We want to use `Method` as part of a config value, deserialized from yaml, in order to easily restrict values to what is supported as HTTP Method by libraries using this lib (e.g. tide and surf).

Currently we have to throw in some custom machinery to make use of the `FromStr` impl on `Method`, which is rather cumbersome (to me).

If nothing speaks against it, I can provide the necessary changes as a PR.

mfelsche:
This fixes #260

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -1,3 +1,5 @@
+use serde::de::{Error as DeError, Unexpected};
+use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt::{self, Display};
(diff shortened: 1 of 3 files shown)


Fishrock123 (approved):

yoshuawuyts (approved):
Oh wow; this is great! -- thanks so much!

PR #263, opened 2020-10-27
crate: add features to switch Error backend

Fishrock123:
Allows end users to choose between [anyhow](https://crates.io/crates/anyhow) and [eyre](https://crates.io/crates/eyre).

If using `eyre`, it is up to the application author to install an eyre reporter themselves.

This is an updated rebase of https://github.com/http-rs/http-types/pull/215, with some notable changes:
- No longer does an eyre reporter install.
- Is all behind feature flags.

A concern: adding `error_anyhow` could break people using `default-features = false`.


Fishrock123:
Upon slightly further investigation - I think Tide may need to hook into this more manually in it's logger (or in any custom logger). More investigation is needed.

PR #265, opened 2020-11-01
Add `cache::ClearSiteData` header

yoshuawuyts:
Adds the `cache::ClearSiteData` header. Ref #99. Thanks!

### DIFF
--- a/src/cache/clear_site_data/directive.rs
+++ b/src/cache/clear_site_data/directive.rs
@@ -1,0 +1,59 @@
+use std::fmt::Display;
+use std::str::FromStr;
+
+/// An HTTP `Clear-Site-Data` directive.
+///
+/// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data#Directives)
+#[derive(Debug, Eq, PartialEq, Hash, Clone, Copy)]
+pub enum ClearDirective {
+    /// Indicates that the server wishes to remove locally cached data (i.e. the
+    /// browser cache, see HTTP caching) for the origin of the response URL.
+    /// Depending on the browser, this might also clear out things like
+    /// pre-rendered pages, script caches, WebGL shader caches, or address bar
+    /// suggestions.
+    Cache,
+    /// Indicates that the server wishes to remove all cookies for the origin of
+    /// the response URL. HTTP authentication credentials are also cleared out.
+    /// This affects the entire registered domain, including subdomains. So
+    /// https://example.com as well as https://stage.example.com, will have
+    /// cookies cleared.
+    Cookies,
+    /// Indicates that the server wishes to remove all DOM storage for the origin
+    /// of the response URL.
+    Storage,
+    /// Indicates that the server wishes to reload all browsing contexts for the
+    /// origin of the response (Location.reload).
+    ExecutionContexts,
+}
+
+impl ClearDirective {
+    /// Get the formatted string.
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            ClearDirective::Cache => r#""cache""#,
+            ClearDirective::Cookies => r#""cookies""#,
+            ClearDirective::Storage => r#""storage""#,
+            ClearDirective::ExecutionContexts => r#""executionContexts""#,
+        }
+    }
+}
+
+impl Display for ClearDirective {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+impl FromStr for ClearDirective {
+    type Err = std::string::ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            r#""cache""# => Ok(Self::Cache),
+            r#""cookies""# => Ok(Self::Cookies),
+            r#""storage""# => Ok(Self::Storage),
+            r#""executionContexts""# => Ok(Self::ExecutionContexts),
+            _ => todo!(),
+        }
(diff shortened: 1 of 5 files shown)


src/cache/clear_site_data/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+pub use directive::ClearDirective;
+
+/// Clear browsing data (cookies, storage, cache) associated with the
+/// requesting website.
+///
+/// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data)
+///
+/// # Specifications
+///
+/// - [RFC 7232, section 3.1: If-Match](https://tools.ietf.org/html/rfc7232#section-3.1)
```
Fishrock123, line 23: This is incorrect but I also cant find an rfc for this? Maybe I don't know how to search for these rfcs...

src/cache/clear_site_data/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        Self {
+            entries: vec![],
+            wildcard: false,
+        }
+    }
+
+    /// Create a new instance from headers.
+    pub fn from_headers(headers: impl AsRef<Headers>) -> crate::Result<Option<Self>> {
+        let mut entries = vec![];
+        let headers = match headers.as_ref().get(CLEAR_SITE_DATA) {
```
Fishrock123, line 64:
```suggestion
        let header_values = match headers.as_ref().get(CLEAR_SITE_DATA) {
```

src/cache/clear_site_data/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+                    continue;
+                }
+                entries.push(ClearDirective::from_str(part)?);
+            }
+        }
+
+        Ok(Some(Self { entries, wildcard }))
+    }
+
+    /// Sets the `If-Match` header.
```
Fishrock123, line 84:
```suggestion
    /// Sets the `Clear-Site-Data` header.
```

src/lib.rs
```diff
@@ -94,7 +94,7 @@
 //! declare it up front. But if you don't, things will still work.
 
 #![deny(missing_debug_implementations, nonstandard_style)]
-#![warn(missing_docs, unreachable_pub)]
```
Fishrock123, line 97: Why this change? It should be `pub(crate)` instead then, no?

src/cache/clear_site_data/mod.rs
jbr: https://w3c.github.io/webappsec-clear-site-data/ — i tend to go to mdn and then check their reference section

src/lib.rs
yoshuawuyts: The lint triggers lots of false positives. We've gone through enabling / disabling it in the past. For whatever reason it was flagging `cache::ClearSiteData` as "private" when it's clearly public, so I think it's time to disable the lint again.

src/cache/clear_site_data/directive.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl ClearDirective {
+    /// Get the formatted string.
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            ClearDirective::Cache => r#""cache""#,
+            ClearDirective::Cookies => r#""cookies""#,
+            ClearDirective::Storage => r#""storage""#,
+            ClearDirective::ExecutionContexts => r#""executionContexts""#,
```
jbr, line 36:
```suggestion
            ClearDirective::Cache => "cache",
            ClearDirective::Cookies => "cookies",
            ClearDirective::Storage => "storage",
            ClearDirective::ExecutionContexts => "executionContexts",
```

Why use rust string syntax when unnecessary?

src/cache/clear_site_data/directive.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl FromStr for ClearDirective {
+    type Err = std::string::ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            r#""cache""# => Ok(Self::Cache),
+            r#""cookies""# => Ok(Self::Cookies),
+            r#""storage""# => Ok(Self::Storage),
+            r#""executionContexts""# => Ok(Self::ExecutionContexts),
```
jbr, line 55:
```suggestion
            "cache" => Ok(Self::Cache),
            "cookies" => Ok(Self::Cookies),
            "storage" => Ok(Self::Storage),
            "executionContexts" => Ok(Self::ExecutionContexts),
```

src/cache/clear_site_data/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        let headers = match headers.as_ref().get(CLEAR_SITE_DATA) {
+            Some(headers) => headers,
+            None => return Ok(None),
+        };
+
+        let mut wildcard = false;
+        for value in headers {
+            for part in value.as_str().trim().split(',') {
+                let part = part.trim();
+                if part == r#""*""# {
```
jbr, line 73:
```suggestion
                if part == "*'"{
```

src/cache/clear_site_data/directive.rs
yoshuawuyts: The quotes here are intentional! Each value **must** include quotes, so it's strings like `"cookies", "storage"`. Hence the usage of raw strings. There's a test for this at the bottom of the file.

jbr: Ohhh! I see. Sorry!

yoshuawuyts:
Updated with all feedback. Merging!

PR #266, opened 2020-11-02
Add `other::Expect` header

yoshuawuyts:
Add `other::Expect` header. Ref #99. Thanks!

### DIFF
--- a/src/other/expect.rs
+++ b/src/other/expect.rs
@@ -1,0 +1,110 @@
+use crate::ensure_eq_status;
+use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, EXPECT};
+
+use std::fmt::Debug;
+use std::option;
+
+/// HTTP `Expect` header
+///
+/// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect)
+///
+/// # Specifications
+///
+/// - [RFC 7231, section 5.1.1: Expect](https://tools.ietf.org/html/rfc7231#section-5.1.1)
+///
+/// # Examples
+///
+/// ```
+/// # fn main() -> http_types::Result<()> {
+/// #
+/// use http_types::Response;
+/// use http_types::other::Expect;
+///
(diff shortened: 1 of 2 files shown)


Fishrock123:
Is `100-continue` the only relevant value here?

yoshuawuyts:
@Fishrock123 yep, and it seems unlikely any others will be added anytime soon. No known browser currently uses this mechanism; it seems curl is the primary user of this right now.

src/other/expect.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// Get the `HeaderName`.
+    pub fn name(&self) -> HeaderName {
+        EXPECT
+    }
+
+    /// Get the `HeaderValue`.
+    pub fn value(&self) -> HeaderValue {
+        let value = "100-continue";
```
Fishrock123, line 71:
```suggestion
        let value = b"100-continue";
```

Fishrock123:
Should we even implement this? Does Curl require it or something? i.e. is it useful?

jbr:
We use it in async-h1, and when async-h1 client behavior is slightly more mature it should also use this header. I don't really see much value in including it as a reusable type, though. It's just as easy to do `.insert_header("Expect", "100-continue")`

yoshuawuyts:
@jbr I mostly included it for completion. We also have typed headers for e.g. `Content-Length` which is arguably trivial too.

Fishrock123:
Parsing `Content-Length` is not exactly uncommon though.

src/other/expect.rs
yoshuawuyts: This requires an additional `to_vec` call, which doesn't seem like much of an improvement over calling `into`.

yoshuawuyts:
It seems the conversation now is about useful this header will turn out to be, rather than whether the implementation is correct. I don't see any concrete downsides in including this; the worst case seems to be that nobody will care. With the upside that this might actually be helpful to some.

Making a decision to merge this.

PR #267, opened 2020-11-02
Add `client_hints::SaveData` header

yoshuawuyts:
Adds the `client_hints::SaveData` header. Ref #99. Thanks!

### DIFF
--- a/src/client_hints/save_data.rs
+++ b/src/client_hints/save_data.rs
@@ -1,0 +1,117 @@
+use crate::bail_status;
+use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, SAVE_DATA};
+
+use std::fmt::Debug;
+use std::option;
+
+/// HTTP `SaveData` header
+///
+/// This header is considered "experimental" and may be subject to change as the
+/// spec evolves.
+///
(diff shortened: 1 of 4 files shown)


src/client_hints/save_data.rs
```diff
@@ -0,0 +1,117 @@
+use crate::bail_status;
+use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, SAVE_DATA};
+
+use std::fmt::Debug;
+use std::option;
+
+/// HTTP `SaveData` header
+///
+/// This header is considered "experimental" and may be subject to change as the
```
Fishrock123, line 9: I think we should put this behind e.g. `"unstable"`...

Fishrock123 (changes requested):

src/client_hints/save_data.rs
yoshuawuyts: Hmm, yeah fair -- or perhaps just not include them for now since they _are_ experimental. I'm so-so about it.

yoshuawuyts:
Decided on Zulip today we should probably postpone the client_hints headers impls until they've stabilized.

PR #268, opened 2020-11-02
Add `content::ContentLocation` header

yoshuawuyts:
Adds the `content::ContentLocation` header. Ref #99. Thanks!

### DIFF
--- a/src/content/content_location.rs
+++ b/src/content/content_location.rs
@@ -1,0 +1,130 @@
+use std::{convert::TryInto, str::FromStr};
+
+use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_LOCATION};
+use crate::url::Url;
+
+/// Indicate an alternate location for the returned data
+///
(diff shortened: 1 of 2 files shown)


jbr:
I'm not familiar with this header, but the spec says this can be a partial-uri, but the current implementation requires it parse to a URL

Fishrock123:
Err, there is already a PR open for this? https://github.com/http-rs/http-types/pull/256

yoshuawuyts:
Oops; I forgot https://github.com/http-rs/http-types/pull/256 was in progress. My bad. Closing!

@jbr yes, accurate -- but we've always made a conscious decision in this lib to never work with partials, but always use fully qualified URLs. Partials are convenient when quickly sending data over the wire, but in practice having a uniform `Url` is much more convenient. `hyperium/http` used partials, and we ended up writing a lot of code to work around that.

Being able to use fully qualified URLs is possible because on the client you'll always know the base URL you're requesting, and on the server we can always trace back the `remote` thanks to the HTTP/1.1 spec. Overall this seems like a better UX than working with enums of "is this/ isn't this a partial" or not doing any kind of parsing at all.

PR #269, opened 2020-11-03
Add `content::ContentType` header

yoshuawuyts:
Adds the `content::ContentType` header. Ref #99. Thanks!


yoshuawuyts:
Closing in favor of #270 which includes content negotation through the `Accept` header and builds on the work from this patch.

PR #270, opened 2020-11-03
Add `content::{Accept, ContentType}` headers

yoshuawuyts:
Adds the `content::{Accept, ContentType}` headers, implementing content negotation for media types. Supersedes https://github.com/http-rs/http-types/pull/269 which only implemented `content::ContentType`. Ref #99. Thanks!

### DIFF
--- a/src/content/accept.rs
+++ b/src/content/accept.rs
@@ -1,6 +1,6 @@
 //! Client header advertising available compression algorithms.
 
-use crate::content::MediaTypeProposal;
+use crate::content::{ContentType, MediaTypeProposal};
 use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, ACCEPT};
 use crate::utils::sort_by_weight;
 use crate::{Error, Mime, StatusCode};
@@ -29,10 +29,10 @@ use std::slice;
 /// accept.push(MediaTypeProposal::new(mime::PLAIN, None)?);
 ///
 /// let mut res = Response::new(200);
-/// let media_type = accept.negotiate(&[mime::XML])?;
-/// media_type.apply(&mut res);
+/// let content_type = accept.negotiate(&[mime::XML])?;
+/// content_type.apply(&mut res);
 ///
-/// assert_eq!(res["Content-Encoding"], "text/html");
+/// assert_eq!(res["Content-Type"], "application/xml;charset=utf-8");
 /// #
 /// # Ok(()) }
 /// ```
@@ -111,14 +111,14 @@ impl Accept {
     /// # Errors
     ///
     /// If no suitable encoding is found, an error with the status of `406` will be returned.
-    pub fn negotiate(&mut self, available: &[Mime]) -> crate::Result<Mime> {
+    pub fn negotiate(&mut self, available: &[Mime]) -> crate::Result<ContentType> {
         // Start by ordering the encodings.
         self.sort();
 
         // Try and find the first encoding that matches.
         for accept in &self.entries {
             if available.contains(&accept) {
-                return Ok(accept.clone().into());
+                return Ok(accept.media_type.clone().into());
             }
         }
 
--- a/src/content/accept.rs
+++ b/src/content/accept.rs
@@ -1,4 +1,4 @@
-//! Client header advertising which media types the client is able to understand
+//! Client header advertising which media types the client is able to understand.
 
 use crate::content::{ContentType, MediaTypeProposal};
 use crate::headers::{HeaderName, HeaderValue, Headers, ToHeaderValues, ACCEPT};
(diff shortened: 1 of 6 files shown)


Fishrock123 (changes requested):
I don't think we should merge this until the API for content-type is flushed out more.

As far as I am aware, there are 4 main uses for `Content-Type`:
- Accept, i.e. negotiation. Covered in this PR.
- Determining file type from http data
- Determining http data from file and/or path
- Determining if a response should be compressed on-demand

The latter 3 are not covered here and are, in my experience, the more  necessary usage.

Here's some 23:00 ideas:
- `ext() -> &str` file extension as string
- `From<Path>` & `from_ext(&str)`
- `compressible() -> bool`

All of these require some kind of in-memory database. I was trying to figure out what structure to use for that but really wasn't sure. However we should populate that database from [mime-db](https://github.com/jshttp/mime-db), since it is by far the most comprehensive source out there (it is an aggregate plus more).

We could also do mime sniffing... but that's kinda costly, requires intercepting bodies, is possibly less reliable, etc

yoshuawuyts:
@Fishrock123 To me it sounds like the features you're suggesting may be out of scope for this PR, even if they seem useful. This PR introduces two typed headers only:

- `Accept` which is used for content negotiation.
- `Content-Type` which is used to communicate the media type used.

Both headers implement the full spec, and in addition to that we also include the content negotiation algorithm to actually make `Accept` _useful_.

Your suggestions around `mime-db` likely best be implemented as extensions to our existing `http_types::mime::Mime` implementation. `Mime` already has some capabilities to infer the type either by reading magic bytes ([`Mime::sniff`](https://docs.rs/http-types/2.7.0/http_types/mime/struct.Mime.html#method.sniff)) or by attempting to guess the mime from the file extension ([`Mime::from_extension`](https://docs.rs/http-types/2.7.0/http_types/mime/struct.Mime.html#method.from_extension)).

We also already have a direct "create a mime from a file" conversion available through ([`Body::from_file`](https://docs.rs/http-types/2.7.0/http_types/struct.Body.html#method.from_file)), which applies both the `from_extension` and `sniff` methods to determine the file type. Finally `Mime` also has a `Display` impl that can be used to convert it back out to a string. Though a `Mime` -> file extension conversion may indeed be useful.

In short: I think your suggestions are good, but they don't need to block this PR. I've filed separate issues to track what's not been implemented so we can remember to implement them later:
- Add `Mime::into_extension`
- Add `Mime::is_compressible`
- Port `mime-db` to Rust to enhance our mime extension guessing
- Meta issue to track media type improvements

Fishrock123:
> I don't think we should merge this until the API for content-type is flushed out more.

Ah, well, this and the issues from #274 are what I meant by this - i.e. that we have at least made an effort to think about it.

src/mime/mod.rs
```diff
@@ -109,6 +109,33 @@ impl Mime {
             })
             .flatten()
     }
+
+    /// Remove a param from the set. Returns the `ParamValue` if it was contained within the set.
+    pub fn remove_param(&mut self, name: impl Into<ParamName>) -> Option<ParamValue> {
```
Fishrock123, line 114: Should this be `take_param`? (Idk.)

src/content/content_type.rs
```diff
@@ -0,0 +1,140 @@
+use std::{convert::TryInto, str::FromStr};
+
+use crate::headers::{HeaderName, HeaderValue, Headers, CONTENT_TYPE};
+use crate::Mime;
+
+/// Indicate the media type of the resource.
```
Fishrock123, line 6:
```suggestion
/// Indicate the media type of a resource's content.
```

src/content/content_type.rs
```diff
@@ -0,0 +1,140 @@
...
+/// # Examples
+///
+/// ```
+/// # fn main() -> http_types::Result<()> {
+/// #
+/// use http_types::content::ContentType;
+/// use http_types::{Response, Mime};
+/// use std::str::FromStr;
+///
+/// let content_type = ContentType::new(Mime::from_str("text/*")?);
```
Fishrock123, line 24: Can this be?:
```suggestion
/// let content_type = ContentType::new("text/*");
```

Or is the issue that `new()` here unwraps?

src/content/accept.rs
```diff
@@ -0,0 +1,433 @@
...
+/// ```
+/// # fn main() -> http_types::Result<()> {
+/// #
+/// use http_types::content::{Accept, MediaTypeProposal};
+/// use http_types::{mime, Response};
+///
+/// let mut accept = Accept::new();
+/// accept.push(MediaTypeProposal::new(mime::HTML, Some(0.8))?);
+/// accept.push(MediaTypeProposal::new(mime::XML, Some(0.4))?);
+/// accept.push(MediaTypeProposal::new(mime::PLAIN, None)?);
```
Fishrock123, line 37:
```suggestion
/// accept.push(mime::PLAIN);
```

Fishrock123 (approved):
Overall looks good, with some minor fixes & questions.

src/mime/mod.rs
yoshuawuyts: This is modeled after [`HashMap::remove`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove) -- it removes the value; however if the entry existed it returns it.

PR #275, opened 2020-11-05
ci improvements

Fishrock123:
- run on "main"
- really check everything
- move to non-deprecated --workspace from --all
- reduce unecessary actions-rs usage
- really clippy everything
- docs without deps

+ run fmt-clippy-docs on beta

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,11 +53,6 @@ jobs:
     steps:
     - uses: actions/checkout@master
 
-   - name: nightly setup
-      run: |
-        rustup toolchain install nightly
-        rustup default nightly
-
     - name: clippy
       run: cargo clippy --workspace --benches --bins --examples --tests --features "hyperium_http,unstable" -- -D warnings
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
       - staging
       - trying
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -32,39 +33,39 @@ jobs:
...
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: nightly
-        components: rustfmt, clippy
-        override: true
+
+   - name: beta setup
+      run: |
+        rustup toolchain install beta
+        rustup default beta
```
yoshuawuyts, line 59: Can you say more why using `beta` makes sense here?

yoshuawuyts: Oh I guess it's just to get us off nightly?

yoshuawuyts (approved):
No objections; just a question

.github/workflows/ci.yaml
Fishrock123: Yeah, nightly is currently missing `rustfmt`...

Fishrock123:
Hmm, right we'd need to actually install clippy and rustfmt for beta. How about we just use stable? it's preinstalled on the github actions images and so it would be much faster. But maybe we want to be ahead of the game with beta? Not 100% sure.

@yoshuawuyts What do you think?

Fishrock123:
Ok, this works now, and has less complexity to just run fmt-clippy-docs on stable, as discussed in zulip.

Merging this and the equivalent PRs on the other repos.

PR #277, opened 2020-11-05
Add `other::Referer` header

yoshuawuyts:
Adds the `other::Referer` header. Ref #99. Thanks!

### DIFF
--- a/src/other/referer.rs
+++ b/src/other/referer.rs
@@ -1,0 +1,162 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, REFERER};
+use crate::{bail_status as bail, Status, Url};
+
+use std::convert::TryInto;
+
+/// Contains the address of the page making the request.
+///
+/// __Important__: Although this header has many innocent uses it can have
+/// undesirable consequences for user security and privacy.
+///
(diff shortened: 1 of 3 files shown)


src/other/referer.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let header_value = headers.iter().last().unwrap();
+
+        let url = match Url::parse(header_value.as_str()) {
+            Ok(url) => url,
+            Err(_) => match base_url.try_into() {
+                Ok(base_url) => base_url.join(header_value.as_str().trim()).status(400)?,
+                Err(_) => bail!(400, "Invalid base url provided"),
```
Fishrock123, line 66: I think this should be 500?

src/other/referer.rs
```diff
@@ -0,0 +1,162 @@
...
+        unsafe { HeaderValue::from_bytes_unchecked(output.into()) }
+    }
+
+    /// Get the url.
+    pub fn location(&self) -> &Url {
+        &self.location
+    }
+
+    /// Set the url.
+    pub fn set_location<U>(&mut self, location: U)
```
Fishrock123, line 97: I think this should return a `Result`.

Fishrock123 (changes requested):

PR #278, opened 2020-11-05
Add `other::SourceMap` header

yoshuawuyts:
Adds the `other::SourceMap` header. Ref #99. Thanks!

### DIFF
--- a/src/headers/constants.rs
+++ b/src/headers/constants.rs
@@ -150,6 +150,9 @@ pub const SERVER: HeaderName = HeaderName::from_lowercase_str("server");
 ///  The `Server` Header
 pub const SERVER_TIMING: HeaderName = HeaderName::from_lowercase_str("server-timing");
 
+///  The `SourceMap` Header
+pub const SOURCE_MAP: HeaderName = HeaderName::from_lowercase_str("sourcemap");
+
 ///  The `Te` Header
 pub const TE: HeaderName = HeaderName::from_lowercase_str("te");
 
--- a/src/other/source_map.rs
+++ b/src/other/source_map.rs
@@ -1,0 +1,159 @@
+use crate::headers::{HeaderName, HeaderValue, Headers, SOURCE_MAP};
+use crate::{bail_status as bail, Status, Url};
+
+use std::convert::TryInto;
+
+/// Links to a file that maps transformed source to the original source.
+///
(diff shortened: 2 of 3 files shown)


src/headers/constants.rs
```diff
@@ -147,6 +147,9 @@ pub const SERVER: HeaderName = HeaderName::from_lowercase_str("server");
 ///  The `Server` Header
 pub const SERVER_TIMING: HeaderName = HeaderName::from_lowercase_str("server-timing");
 
+///  The `SourceMap` Header
+pub const SOURCE_MAP: HeaderName = HeaderName::from_lowercase_str("sourcemap");
```
Fishrock123, line 151: Should this be `SOURCEMAP`?

src/other/source_map.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        // If we successfully parsed the header then there's always at least one
+        // entry. We want the last entry.
+        let header_value = headers.iter().last().unwrap();
+
+        let url = match Url::parse(header_value.as_str()) {
+            Ok(url) => url,
+            Err(_) => match base_url.try_into() {
+                Ok(base_url) => base_url.join(header_value.as_str().trim()).status(400)?,
+                Err(_) => bail!(400, "Invalid base url provided"),
```
Fishrock123, line 63: Also should probably be 500?

src/other/source_map.rs
```diff
@@ -0,0 +1,159 @@
...
+        unsafe { HeaderValue::from_bytes_unchecked(output.into()) }
+    }
+
+    /// Get the url.
+    pub fn location(&self) -> &Url {
+        &self.location
+    }
+
+    /// Set the url.
+    pub fn set_location<U>(&mut self, location: U)
```
Fishrock123, line 94: Same as `Referrer`, I think this should return `Result`.

Fishrock123 (dismissed):

src/other/source_map.rs
jbr: or probably just not set a status explicitly

src/headers/constants.rs
yoshuawuyts: The canonical spelling of the header is `SourceMap` -- which is not quite `Source-Map` but I think `SOURCE_MAP` is the right capitalization.

PR #279, opened 2020-11-08
Implement `ToHeaderValues` for more string types

yoshuawuyts:
Implements `ToHeaderValues` for more string types. Noticed that passing `&format!("{}", "value")` as the header value didn't work, but `format!("{}", "value").as_str()` does. This fixes that. Thanks!

### DIFF
--- a/src/headers/to_header_values.rs
+++ b/src/headers/to_header_values.rs
@@ -54,10 +54,7 @@ impl ToHeaderValues for String {
     type Iter = option::IntoIter<HeaderValue>;
 
     fn to_header_values(&self) -> crate::Result<Self::Iter> {
-        let value = self
-            .parse()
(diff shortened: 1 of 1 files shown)


Fishrock123 (approved):

jbr (commented):
Could these be defined in terms of the `ToHeaderValues for &'a str` impl to reduce duplicated code? Similarly, the String one probably could also be defined as 

```rust
impl ToHeaderValues for String {
    type Iter = option::IntoIter<HeaderValue>;

    fn to_header_values(&self) -> crate::Result<Self::Iter> {
        self.as_str().to_header_values()
    }
}
```

That way if changes were needed to the &str impl, it wouldn't have to be repeated in four places

PR #282, opened 2020-11-15
do not let Body read beyond its length

jbr:
this logic is currently handled in several places in async-h1, but it should be the responsibility of Body to ensure that if it has a set length, it will not read beyond that length.

I believe this is semver-patch, unless we consider "reading past set body length" intended behavior

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,4 +1,4 @@
-use futures_lite::{io, prelude::*};
+use futures_lite::{io, prelude::*, ready};
 use serde::{de::DeserializeOwned, Serialize};
 
 use std::fmt::{self, Debug};
@@ -56,6 +56,7 @@ pin_project_lite::pin_project! {
         reader: Box<dyn AsyncBufRead + Unpin + Send + Sync + 'static>,
         mime: Mime,
         length: Option<usize>,
+        bytes_read: usize
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/body.rs
```diff
@@ -459,15 +468,25 @@ impl AsyncRead for Body {
         cx: &mut Context<'_>,
         buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
-        Pin::new(&mut self.reader).poll_read(cx, buf)
+        let mut buf = match self.length {
+            None => buf,
+            Some(length) if length == self.bytes_read => return Poll::Ready(Ok(0)),
```
jbr, line 473: this line is an optional optimization and could be removed if it makes the code more readable to do so. i was on the fence about this tradeoff

src/body.rs
```diff
@@ -459,15 +468,25 @@ impl AsyncRead for Body {
...
+            Some(length) => {
+                let max_len = (length - self.bytes_read).min(buf.len());
+                &mut buf[0..max_len]
```
Fishrock123, line 476: Will this panic if the body has more bytes than it specifies?

jbr: nope, there's a test for that. we just won't read them from the underlying Read, which is the correct behavior for keepalive

jbr: this line limits the length of the mutable slice that we pass into the inner read implementation, effectively telling it "give me a maximum of `max_len` bytes

PR #287, opened 2020-11-25
make it so upgrade connections can be built from the outside

jbr:


src/upgrade/connection.rs
```diff
@@ -9,6 +9,12 @@ pub struct RawConnection<Inner> {
     inner: Inner,
```
Fishrock123, line 9: If `Connection` is `RawConnection<Box<dyn T>>` anyways, why isn't `inner just directly `Box<dyn T>`?

Fishrock123 (commented):
This structure strikes me as being a bit odd, see comment below.

However, along the lines of this PR, might we also want a `From<Box<dyn T>>`?

jbr:
This will only ever be used in one place probably, so it's not worth adding an Into. I think this code probably could be simplified, but that's not what this PR is about

Fishrock123 (approved):

src/upgrade/connection.rs
yoshuawuyts: Good question; not sure. I think we should pick that up when stabilizing the `upgrade`submodule.

yoshuawuyts (approved):

PR #290, opened 2020-11-27
Fix [sic] in docs

yoshuawuyts:
This was causing a link breakage warning in the docs. Thanks!


Fishrock123:
tbh this should probably link to https://en.wikipedia.org/wiki/Sic because not everyone will know what it means even though it is a good use of it.

PR #291, opened 2020-11-27
Stabilize upgrade

yoshuawuyts:
Proposes to stabilize the `upgrade` submodule. I've included the changes suggested in https://github.com/http-rs/http-types/pull/287#discussion_r530688787, simplifying the internals. One downside of this, however, is that we remove the `Clone` derive. I'm not sure if this leads to a loss of functionality though, since derives are conditional, and we're now always boxing.

@jbr giving this a test run with your websockets tide patch would be much appreciated. That's what this API was designed for, and it allows us to gage whether this actually works as expected. Thanks!

### DIFF
--- a/src/headers/constants.rs
+++ b/src/headers/constants.rs
@@ -135,6 +135,9 @@ pub const PROXY_AUTHENTICATE: HeaderName = HeaderName::from_lowercase_str("proxy
 ///  The `Proxy-Authorization` Header
 pub const PROXY_AUTHORIZATION: HeaderName = HeaderName::from_lowercase_str("proxy-authorization");
 
+/// The `Proxy-Connection` Header
+pub const PROXY_CONNECTION: HeaderName = HeaderName::from_lowercase_str("proxy-connection");
+
 ///  The `Referer` Header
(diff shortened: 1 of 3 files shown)


Fishrock123:
If cloning is desirable, we could use [dyn-clone](https://crates.io/crates/dyn-clone).

Fishrock123:
To be fair, I don't think that `Clone` would have worked before anyways unless you had something that wasn't `Connection`.

jbr:
this branch works with websockets 👍
both tide-websockets and [the demo chat app](https://github.com/jbr/littoral) are now using it with no changes

jbr (changes requested):
response.rs still treats upgrades as unstable

jbr:
to check if this branch integrates with the upgrades branch of async-h1, rebuild here: https://github.com/http-rs/async-h1/pull/159/checks?check_run_id=1465731580 — the current failure represents the missing changes in response.rs

yoshuawuyts:
Annnnnnd https://github.com/http-rs/async-h1/pull/159/checks?check_run_id=1500647266 is passing! -- I think that means we can merge this!

Thanks for the review everyone, and thanks @pepoviola for the patch fix!

Merging!

PR #292, opened 2020-11-30
stabilize upgrade in response

pepoviola:
Hi @yoshuawuyts / @jbr, I start working in an example using websocket based on [littoral](https://github.com/jbr/littoral) and I found that some changes are needed in this branch `stabilize-upgrade` to support `ws`.

I tested ok in my example but let me know if something is wrong, thanks for the awesome job!!! 🙌

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -15,10 +15,7 @@ use crate::headers::{
 use crate::mime::Mime;
 use crate::trailers::{self, Trailers};
 use crate::{Body, Extensions, StatusCode, Version};
-
-cfg_unstable! {
-    use crate::upgrade;
-}
+use crate::upgrade;
 
 #[cfg(not(feature = "unstable"))]
 pin_project_lite::pin_project! {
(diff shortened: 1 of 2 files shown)


src/response.rs
```diff
@@ -91,31 +56,6 @@ impl Response {
     /// Create a new response.
     #[cfg(not(feature = "unstable"))]
```
jbr, line 57: is this line still needed?

src/response.rs
```diff
@@ -15,10 +15,7 @@ use crate::headers::{
 use crate::mime::Mime;
 use crate::trailers::{self, Trailers};
 use crate::{Body, Extensions, StatusCode, Version};
-
-cfg_unstable! {
-    use crate::upgrade;
-}
+use crate::upgrade;
 
 #[cfg(not(feature = "unstable"))]
```
jbr, line 20: is this line still needed?

jbr (changes requested):
looks like cfg_unstable needs to be removed from src/utils/mod.rs for ci to pass, and we should check if we're actually using the unstable feature anywhere. if not, we should remove the feature

src/response.rs
pepoviola: I don't think so because there isn't any unstable feature anymore. I will remove those. Thx!

src/response.rs
pepoviola: removed 😄

jbr (approved):

jbr:
refs: #291

Fishrock123 (approved):
https://github.com/http-rs/http-types/pull/291 should come first but if we are stabilizing that then this goes hand-in-hand.

jbr:
#291 doesn't work without this, and this is a pr into #291. There is no reason to merge #291 without this

Fishrock123:
Oh, guess i had it backwards then 😅

yoshuawuyts (approved):
Thank you!

PR #295, opened 2020-12-08
Change Mime to MediaType

brightly-salty:
This is a WIP: will only be merged after other semver-major changes, and will likely have conflicts then.

This PR renames the `mime` module to `media_type`, the `Mime` struct to `MediaType`, any associated functions with "mime" in their name to "media_type", any `mime` fields to `media_type`, and all related docs have also changed.

This is my first PR for http-types. Let me know if anything else is needed of me.


yoshuawuyts:
@brightly-salty thanks so much for this PR. We'll indeed have to wait to merge this until we're ready for a 3.0 (which may be soon-ish actually). But it appears some of the comments have been caught in the diff too: "Mime type" in docs should probably be rename to "media type". Currently some examples appear to be "MediaType type" and "media_type type". If you could do another pass to update the comments, that'd be :100:

yoshuawuyts (approved):

yoshuawuyts:
Converted to draft so we don't actually merge it. But there already appear to be merge conflicts. These don't need to be resolved straight away, but it's probably worth pointing out regardless.

In terms of timing for an http-types 3.0 release; we're having a meeting early next year with the core team to determine the schedule for that. Having all this work already ready to be merged is a huge boon and gives us confidence we'll be able to make this release happen sooner rather than later. Thank you!

PR #298, opened 2020-12-14
construct bodies from impl Read instead of impl BufRead

### LINKED ISSUE #297: Body should be constructed around a Read instead of a BufRead
this will require dropping the BufRead implementation for Body, Request, and Response as well as changing the Body::from_reader signature — people can wrap Reads with async_std::io::BufReaders if they need BufRead behavior

jbr:
closes #297

do not merge until 3.0


yoshuawuyts:
Marked as draft so we physically can't merge until 3.0.

src/body.rs
```diff
@@ -1,4 +1,4 @@
-use futures_lite::{io, prelude::*, ready};
+use futures_lite::{io, prelude::*, ready, AsyncRead as Read};
```
yoshuawuyts, line 1: We should probably keep calling this `AsyncRead`. I think that's a naming mistake we made with `async-std`.

```suggestion
use futures_lite::{io, prelude::*, ready, AsyncRead};
```

yoshuawuyts (approved):
One nit, but otherwise good!

Fishrock123:
What's this waiting on?

jbr:
I think this one should probably be a draft. I'm not certain this is the right choice for async-h1, but can throw together a PR branch for async-h1 that uses this branch and see what the consequences are

PR #299, opened 2020-12-18
Upgrade pin-project-lite to 0.2.0

joshtriplett:
Ran into this when looking for duplicate crate dependencies with
different versions.

I've confirmed that this change builds and passes all tests.


yoshuawuyts (approved):
Fantastic; thanks so much!

PR #304, opened 2020-12-19
Fix formatting for StatusCode

### LINKED ISSUE #302: Change StatusCode Display and Debug impls to include canonical representations
The output is currently `500` for `Display` and `InternalServerError` for `Debug`. In both cases it would be better to display `500: Internal Server Error`.

brightly-salty:
Fix formatting for StatusCode
(also ran cargo format)

Fixes #302.


joshtriplett:
Is this actually a semver-major change? It just changes the `Debug` output, which isn't normally considered a breaking change.

PR #305, opened 2020-12-19
Remove unneeded re-exports

### LINKED ISSUE #264: Remove http header re-exports from crate root
We currently export the following headers from the crate root:

Screenshot_2020-10-31 http_types - Rust(2)

A majority of these are re-exports from our sub-modules. As our index is growing because of the addition of a variety of typed headers I would like to trim down some of these re-exports and leave them accessible only as part of the sub-modules. The types to keep would be only the types required to construct a request / response, or types that don't currently already have a dedicated submodule:

__keep__
- `Body` core type
- `Error` core type
- `Extensions` no dedicated submodule
- `Request` core type
- `Response` core type
- `Url` core type (for now; revisit after the `TryFrom` patch on `url` is published)

__move to submodule__
- `Cookie` has submodule, not critical
- `Headers` rarely referenced directly, and when used it's in conjunction with other `header` types
- `Mime` (rename to `MediaType` (see #251 )
- `Trailers` has a submodule and uncommonly used

brightly-salty:
Remove unneded reexports from crate root
Update documentation

Fixes #264


joshtriplett:
Manually merged to fix merge conflicts. Thank you!

PR #306, opened 2020-12-19
Allow omitting cookie support, with a new "cookies" feature

joshtriplett:
The "cookie" crate introduces a substantial dependency tree. Allow
people to omit that dependency if they don't need it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,8 @@ docs = ["unstable"]
 unstable = []
 hyperium_http = ["http"]
 async_std = ["fs"]
-cookie-secure = ["cookie/secure"]
+cookies = ["cookie"]
+cookie-secure = ["cookies", "cookie/secure"]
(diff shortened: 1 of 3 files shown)


joshtriplett:
After this, I'd like to change async-h1 to use `default-features = false`, and change tide to only depend on `http-types/cookies` if its own `cookies` feature is set.

yoshuawuyts (approved):
This seems reasonable, thank you!

joshtriplett:
@yoshuawuyts Thanks! Would it be possible to get a new http-types upload, to allow tide to make use of this change?

yoshuawuyts:
We do monthly releases for all the main http-rs libraries, with the next release of http-types planned for the week of January 10th.

Usually I wouldn't mind doing a patch release if requested, but I'm actually planning to spend the holidays away from the computer. Hopefully that's not too late!

PR #307, opened 2020-12-19
Make MIME tests conditional on the "fs" feature

joshtriplett:
This allows `cargo test --no-default-features` to pass, which helps when
testing if the code is functional without a feature.


yoshuawuyts (approved):
This looks great; thanks heaps!

PR #308, opened 2020-12-19
Add error handling to Request methods

### LINKED ISSUE #303: Return `Result` from constructors which take `TryInto<Url>`
Now that `TryFrom<str> for Url` has been implemented, using the `Url` struct is a lot more pleasant. This opens up the possibility again for us to move from a "panic if parsing fails" to returning errors if parsing fails.

```rust
// current, panic if the url is malformed
let req = Request::post("https://api.foo.com/berries");

// proposed, throw an error instead
let req = Request::post("https://api.foo.com/berries")?;
```

brightly-salty:
Make all associated Request methods return a Result<Self, Error> instead of Self and panicking.
Change docs and doc-tests to reflect this, and remove manual parsing in doc-tests and unit tests

Fixes #303

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -13,7 +13,7 @@ use crate::headers::{
 };
 use crate::mime::Mime;
 use crate::trailers::{self, Trailers};
-use crate::{Body, Extensions, Method, StatusCode, Url, Version};
+use crate::{Body, Error, Extensions, Method, StatusCode, Url, Version};
 
 pin_project_lite::pin_project! {
     /// An HTTP request.
@@ -21,9 +21,9 @@ pin_project_lite::pin_project! {
     /// # Examples
     ///
     /// ```
-    /// use http_types::{Url, Method, Request};
+    /// use http_types::{Method, Request};
     ///
-    /// let mut req = Request::new(Method::Get, Url::parse("https://example.com").unwrap());
+    /// let mut req = Request::new(Method::Get, "https://example.com").unwrap();
     /// req.set_body("Hello, Nori!");
     /// ```
     #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


src/request.rs
```diff
@@ -45,14 +45,14 @@ pin_project_lite::pin_project! {
...
     /// Create a new request.
-    pub fn new<U>(method: Method, url: U) -> Self
+    pub fn new<U>(method: Method, url: U) -> Result<Self, Error>
     where
         U: TryInto<Url>,
-        U::Error: std::fmt::Debug,
+        U::Error: std::fmt::Debug + Into<anyhow::Error>,
     {
-        let url = url.try_into().expect("Could not convert into a valid url");
+        let url = url.try_into().map_err(|e| Error::new(500, e))?;
```
yoshuawuyts, line 53: This does not need the `map_err` call to work:
```suggestion
        let url = url.try_into()?;
```
But if it's for clarity purposes, it's preferable to use the `Status` trait instead:
```suggestion
        let url = url.try_into().status(500)?;
```

src/request.rs
```diff
@@ -682,14 +680,14 @@ impl Request {
...
+    /// let mut req = Request::get("https://example.com").unwrap();
     /// req.set_body("Hello, Nori!");
     /// assert_eq!(req.method(), Method::Get);
     /// ```
-    pub fn get<U>(url: U) -> Self
+    pub fn get<U>(url: U) -> Result<Self, Error>
     where
         U: TryInto<Url>,
-        U::Error: std::fmt::Debug,
+        U::Error: std::fmt::Debug + Into<anyhow::Error>,
```
yoshuawuyts, line 690: All of these calls should instead use `http_types::Error`. Or even better: `http_types::Result`:

```suggestion
    pub fn get<U>(url: U) -> crate::Result<Self>
    where
        U: TryInto<Url>,
        U::Error: std::fmt::Debug,

**Review by yoshuawuyts — CHANGES_REQUESTED** [review]
Few nits, but overall this is a really promising PR. Thank you!

File: src/request.rs (hunk 1 above)
Comment (brightly-salty) [reply src/request.rs:53]: I got the following error after replacing map_err with status

```
no method named `status` found for enum `std::result::Result<url::Url, <U as std::convert::TryInto<url::Url>>::Error>` in the current scope

method not found in `std::result::Result<url::Url, <U as std::convert::TryInto<url::Url>>::Error>`

note: the method `status` exists but the following trait bounds were not satisfied:
      `<U as std::convert::TryInto<url::Url>>::Error: std::error::Error`
      which is required by `std::result::Result<url::Url, <U as std::convert::TryInto<url::Url>>::Error>: status::Status<url::Url, <U as std::convert::TryInto<url::Url>>::Error>`
      `<U as std::convert::TryInto<url::Url>>::Error: std::marker::Send`
      which is required by `std::result::Result<url::Url, <U as std::convert::TryInto<url::Url>>::Error>: status::Status<url::Url, <U as std::convert::TryInto<url::Url>>::Error>`
      `<U as std::convert::TryInto<url::Url>>::Error: std::marker::Sync`
      which is required by `std::result::Result<url::Url, <U as std::convert::TryInto<url::Url>>::Error>: status::Status<url::Url, <U as std::convert::TryInto<url::Url>>::Error>`
```

PR #310, opened 2020-12-24
Refactor media type internals

yoshuawuyts:
This dedupes some internal logic of our media types, and refactors them to use `Cow`. This also moves utf-8 detection from a nested enum to a flag. Overall this simplifies the internals of our media types without any external API changes.

I was trying to address https://github.com/http-rs/tide/issues/27#issuecomment-750921221, but this is still blocked on the stdlib. However it seems that the only change we need to make now is to add `StructuralEq` to `Vec` and `Cow` so that they work in const contexts. Which should be slightly more feasible.

### DIFF
--- a/src/mime/constants.rs
+++ b/src/mime/constants.rs
@@ -1,5 +1,6 @@
 use super::ParamKind;
 use crate::Mime;
+use std::borrow::Cow;
 
 macro_rules! utf8_mime_const {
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
This has been in review for a good while now. This is just a refactor and doesn't add any new API surface, so going to go ahead and merge this.

Issue #313, opened 2020-12-26
Use `ascii` crate for string values inside `HeaderValue`, `HeaderName` types


yoshuawuyts:
We could remove almost every instance of `unsafe` in our codebase by replacing the use of `String` in `HeaderName` and `HeaderValue` with [`ascii::AsciiString`](https://docs.rs/ascii/1.0.0/ascii/) instead. Implementing this shouldn't be too hard, but code-wise it's a fair few lines that need changing.

This wouldn't change anything in our API signatures either, assuming we don't have any bugs in how we handle ASCII today. @brightly-salty would you be interested in perhaps trying to tackle this?

brightly-salty:
Sure!

brightly-salty:
Would you like me to use `from_ascii(..).unwrap()` or `unsafe { from_ascii_unchecked(..)}` It seems like both would work equally well but the second preserves `unsafe`.

yoshuawuyts:
Oh, good question. Yeah, I actually care less about the perf here than I care about removing `unsafe`. I think the former is probably desirable.

jbr:
If we panic on improper ascii, doesn't that mean a http client could shut down the server?

brightly-salty:
We are already calling `ensure!` before the unsafe code. I'm not aware of the specifics of the implementation of `ensure` but I think it will panic in some form. The `unwrap` panicking should be impossible because we are already ensuring it is ASCII beforehand.

brightly-salty:
Should the behavior of `HeaderName::from_lowecase_str` be to panic, be unsafe, or to return a result? Currently it assumes the given `str` is valid ASCII and is lowercase.

brightly-salty:
So the question for `HeaderName::from_lowecase_str` is how to make it work with existing usages. We can use an `unsafe {AsciiStr::from_ascii_unchecked(s)}` but this function is not `const` which we need for `HeaderName::from_lowercase_str` to be `const`, which we need to be able to create `const`s in `src/headers/constants.rs`.

The only solution I can think of is to use lazy_static for the constants, thus allowing the use of non-const functions in their initialization, thus allowing us to use `unsafe {AsciiStr::from_ascii_unchecked(s)}` in `HeaderName::from_lowercase_str`.

@jbr @yoshuawuyts

brightly-salty:
@yoshuawuyts I'm going to "unassign" myself from this one, because it requires multiple high-level design decisions.

PR #315, opened 2020-12-27
Implement the `Header` trait

yoshuawuyts:
_Closes #286_

This PR implements the `Header` trait, creating a unified interface for the convention we had of `TypedHeader::{value, name, apply}`. This instead enables an interface of:

```rust
pub trait Header {
    fn header_name(&self) -> HeaderName;
    fn header_value(&self) -> HeaderValue;
    fn apply_header<H: AsMut<Headers>>(&self, mut headers: H);
}
```

Because `header_name` and `header_value` have to be implemented, `apply_header` has a default implementation. This should make it easier to keep up to date.

## Breaking changes

This patch doesn't break anything, but it does set us up for the 3.0.0 release. When we're ready to implement the 3.0.0 changes we should remove the `name`, `value` and `apply` methods in favor of keeping the trait only. This patch also explicitly chose to keep the `value` method implementation where it is in order to not create a massive merge conflict with #313.

## Bug fixes
This patch also fixes two bugs: a wrong header name in `TE` and in `TransferEncoding`. These have been adjusted and now use the right header name.

## Prior work
This patch serves as an alternative to https://github.com/http-rs/http-types/pull/285, implementing the design I referenced in https://github.com/http-rs/http-types/pull/285#issuecomment-750891612.

## Future directions
We should consider taking a look at the `headers` submodule in general. A few changes I think would be worthwhile:
- Move the header name constants from the crate root to be associated constants on `HeaderName` instead.
- Rename `Headers` to `HeaderMap`, matching the `hyperium/hyper` naming convention
- Consider changing the return type of `ToHeaderValues` from an associated iterator to be an instance of `HeaderValues` instead. The trait precedes the concrete type, and having a trait and a type with similar names be completely unrelated feels awkward.
- Do a pass on the typed headers to ensure the right `AsRef` and `AsMut` implementations are provided.
- Do a pass on the typed headers to clean up the examples, making them practical.
- Do a pass on the typed headers to link each header to their respective MDN documentation.
- Implement per-method examples on the typed headers.
- Create `http_types::headers` crate-level docs with an explainer of the different types in the crate.

### DIFF
--- a/src/auth/authorization.rs
+++ b/src/auth/authorization.rs
@@ -1,6 +1,6 @@
 use crate::auth::AuthenticationScheme;
 use crate::bail_status as bail;
-use crate::headers::{HeaderName, HeaderValue, Headers, AUTHORIZATION};
+use crate::headers::{Header, HeaderName, HeaderValue, Headers, AUTHORIZATION};
 
 /// Credentials to authenticate a user agent with a server.
 ///
@@ -78,7 +78,7 @@ impl Authorization {
 
     /// Get the `HeaderName`.
     pub fn name(&self) -> HeaderName {
-        AUTHORIZATION
+        self.header_name()
     }
 
     /// Get the `HeaderValue`.
@@ -107,6 +107,16 @@ impl Authorization {
     /// Set the authorization credentials.
     pub fn set_credentials(&mut self, credentials: String) {
         self.credentials = credentials;
+    }
+}
+
+impl Header for Authorization {
+    fn header_name(&self) -> HeaderName {
+        AUTHORIZATION
+    }
+
+    fn header_value(&self) -> HeaderValue {
+        self.value()
     }
 }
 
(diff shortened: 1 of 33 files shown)


Fishrock123:
Can `(&str, &str)` implement this `Header` trait?

yoshuawuyts:
@Fishrock123 good question; will have to try. I suspect there may actually be limits on what we can do because of how we're using generics.

yoshuawuyts:
Oh woah, the impl actually worked! Added a test for it too:

```rust
#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn header_from_strings() {
        let strings = ("Content-Length", "12");
        assert_eq!(strings.header_name(), "Content-Length");
        assert_eq!(strings.header_value(), "12");
    }
}
```

yoshuawuyts:
It's been a while. Assuming this is good -- going to merge it.

Fishrock123:
Sorry. I think this should probably be able to deal with borrows if possible, see #336/#335

PR #296, opened 2020-12-09
Rename Body::from_file, add Body::from_file

brightly-salty:
This PR renames `Body::from_file` to `Body::from_path`, and adds `Body::from_file`, which takes an `async_std::fs::File` like the name would suggest.

Fixes #185.

Semver-major.

PR #318, opened 2021-01-15
Rename Body::from_file to Body::from_path, add Body::from_file, handle filename-based MIME-type guessing for File objects

sunfishcode:
This splits a `Body::from_open_file` out from `Body::from_file` so that
users who want to use a file opened in a manner other than `File::open`
can do so.

In particular, this will help users using [`cap_std::fs::Dir::open`] to
open files without requiring http-types to have a dependency on `cap_std`.

[`cap_std::fs::Dir::open`]: https://docs.rs/cap-std/latest/cap_std/fs/struct.Dir.html#method.open

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -349,7 +349,7 @@ impl Body {
         Ok(serde_urlencoded::from_str(&s).status(StatusCode::UnprocessableEntity)?)
     }
 
-    /// Create a `Body` from a file.
+    /// Create a `Body` from a file named by a path.
     ///
     /// The Mime type is sniffed from the file contents if possible, otherwise
     /// it is inferred from the path's extension if possible, otherwise is set
@@ -362,17 +362,17 @@ impl Body {
     /// use http_types::{Body, Response, StatusCode};
     ///
     /// let mut res = Response::new(StatusCode::Ok);
-    /// res.set_body(Body::from_file("/path/to/file").await?);
+    /// res.set_body(Body::from_path("/path/to/file").await?);
     /// # Ok(()) }) }
     /// ```
     #[cfg(all(feature = "fs", not(target_os = "unknown")))]
-    pub async fn from_file<P>(path: P) -> io::Result<Self>
+    pub async fn from_path<P>(path: P) -> io::Result<Self>
     where
         P: AsRef<std::path::Path>,
     {
         let path = path.as_ref();
         let file = async_std::fs::File::open(path).await?;
-        Self::from_open_file(file, path).await
+        Self::from_file(file, path).await
     }
 
     /// Create a `Body` from an already-open file.
@@ -393,11 +393,11 @@ impl Body {
     /// let mut res = Response::new(StatusCode::Ok);
     /// let path = std::path::Path::new("/path/to/file");
     /// let file = async_std::fs::File::open(path).await?;
-    /// res.set_body(Body::from_open_file(file, path).await?);
+    /// res.set_body(Body::from_file(file, path).await?);
     /// # Ok(()) }) }
     /// ```
     #[cfg(all(feature = "fs", not(target_os = "unknown")))]
-    pub async fn from_open_file(
+    pub async fn from_file(
         mut file: async_std::fs::File,
         path: &std::path::Path,
     ) -> io::Result<Self> {
--- a/src/body.rs
+++ b/src/body.rs
@@ -351,8 +351,9 @@ impl Body {
 
     /// Create a `Body` from a file.
     ///
-    /// The Mime type set to `application/octet-stream` if no other mime type has
-    /// been set or can be sniffed.
+    /// The Mime type is sniffed from the file contents if possible, otherwise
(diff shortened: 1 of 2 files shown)


sunfishcode:
The clippy error doesn't look related to this patch.

sunfishcode:
I just noticed that this PR is very similar to another existing PR, https://github.com/http-rs/http-types/pull/296.

The main difference is that this PR's `File` constructor also takes a `Path` for Mime type detection (it may be empty if no path is available).

Also, this PR uses different names: `from_file` and `from_open_file` instead of `from_path` and `from_file`. Either way seems fine to me. The names in this PR are semver-compatible, for what that's worth.

Fishrock123:
I prefer #296. `from_file` should take an already open file. It is probably the developer's responsibility to then ensure an appropriate mime fallback exists if magic byte sniffing doesn't work out.

Fishrock123 (dismissed):
As per my former comment

sunfishcode:
I'm ok with `from_file` taking an already open file.

Concerning the mime type, there are some situations where `mime_guess` will positively identify a file as `application/octet-stream`, in which case checking for `mime::BYTE_STREAM` and overriding it with a fallback would produce different results. Consequently, I'd prefer to keep the `path` argument.

sunfishcode:
I've now updated this PR to use the `from_file`/`from_path` renaming. And rebased it to pick up the clippy fix on main, so the CI is now green. As mentioned above, I preserved the `path` argument in `from_file`.

yoshuawuyts:
If I'm not mistaken this appears to be a breaking change. I'm okay with merging this, but we'll have to do this as part of our 3.0 effort. Luckily that should be kicking off tomorrow (last pre-3.0 http-types release today), so we'll be able to merge this soon! (:

src/body.rs
```diff
@@ -361,16 +362,45 @@ impl Body {
...
+    }
+
+    /// Create a `Body` from an already-open file.
+    ///
+    /// The Mime type is sniffed from the file contents if possible, otherwise
+    /// it is inferred from the path's extension if possible, otherwise is set
+    /// to `application/octet-stream`.
+    ///
+    /// The path here is only used to provide an extension for guessing the Mime
+    /// type, and may be empty if the path is unknown.
```
joshtriplett, line 385: I think I'd prefer `Option<&std::path::Path>` rather than requiring the caller to construct and pass an empty `Path`.

joshtriplett:
Reading this change and https://github.com/http-rs/http-types/pull/296, it looks like the main difference is that https://github.com/http-rs/http-types/pull/296 only tries to guess the MIME type in `from_path`, while this change makes `from_file` take a path that can be used for extension-based MIME-type guessing.

I made one comment above about the path argument (using `Option`).

I personally think the common case of `from_file` will not be to pass a `path`, and I don't think we should make most callers of `from_file` pass that argument. I would propose one of two options, depending on how common you think this is:

- Make this an internal function `from_file_with_path`, and have the public `from_file` pass `None`. This would require callers to specify the MIME type themselves. Callers can call `Mime::from_extension` if they want to do filename-based MIME-type guessing. This seems like the right solution if we don't expect it to be especially common to do filename-based MIME-type guessing for an already-open `File`.
- Make this a public function `from_file_with_path`, and have a public `from_file` that doesn't take a path. This would allow callers to call `from_file_from_path` in the case where they want filename-based MIME-type guessing, or `from_file` if they don't. This seems like the right solution if we expect it to be especially common to do filename-based MIME-type guessing for an already-open `File`.

Personally, I suspect it won't be especially common to pass an already-open `File` but still want filename-based MIME-type guessing, so I would propose that we go with the first option.

Does that seem reasonable to you, @sunfishcode?

joshtriplett:
Also, if we end up merging this one instead of https://github.com/http-rs/http-types/pull/296 , then (assuming you copied the doc comments from https://github.com/http-rs/http-types/pull/296 ) could you please add a Co-authored-by to that one commit? We can't do anything about the duplicate efforts at this point, but I want to make sure folks get appropriate credit.

sunfishcode:
@joshtriplett The `from_file_with_path` option works for me. I don't know how common it will be to pass a path, but my own use case does want this :smile: .

The doc comment wasn't copied from #296; it's likely that we both copied the comment from the pre-existing `from_file` function and made similar obvious changes. That said, I've now added Co-authored-by, to reflect that we both did similar work here.

joshtriplett:
@sunfishcode wrote:
> @joshtriplett The from_file_with_path option works for me. I don't know how common it will be to pass a path, but my own use case does want this :smile:.

Fair enough.

> The doc comment wasn't copied from #296; it's likely that we both copied the comment from the pre-existing from_file function and made similar obvious changes.

Ah, makes sense.

> That said, I've now added Co-authored-by, to reflect that we both did similar work here.

Thank you for your graceful handling of this. I know that it's frustrating when work gets duplicated.

Merging.

PR #320, opened 2021-01-16
expose Headers::new and Extensions::new

jbr:
Other users than http-rs might want to use these types, but they cannot currently be constructed outside of this crate


Fishrock123:
needs a rebase for clippy

yoshuawuyts:
I'm okay with this change, but think we should document that this is most likely not what you want to do. Someone coming from `hyperium/http` may incorrectly assume that `Headers` instances need to manually be created. Instead this really only serves as an intermediate type that is meant to be accessed through `Request`, `Response`, etc. Much like a trait is, but slightly different.

Just exposing this without guidance may cause confusion, and we should try and evade that if possible. (Though it'd be correct to point out that our current guidance on these types isn't great either, and folks may already be confused. But ideally we shouldn't add to that 😅 )

yoshuawuyts:
Oh woah, the impl actually worked! Added a test for it too:

```rust
#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn header_from_strings() {
        let strings = ("Content-Length", "12");
        assert_eq!(strings.header_name(), "Content-Length");
        assert_eq!(strings.header_value(), "12");
    }
}
```

Fishrock123:
I think this could be useful if you could set the headers when constructing a request, at least that would make per-client headers in Surf config a tad nicer under the hood.

CfirTsabari:
> I think this could be useful if you could set the headers when constructing a request, at least that would make per-client headers in Surf config a tad nicer under the hood.

why not using `Middleware`?

Fishrock123:
Unimportant implementation detail, but feature wise, convenience.

Issue #331, opened 2021-02-06
The `Method` enum should support all the registered HTTP methods


zenekron:
**TL;DR**: the `Method` enum only supports fraction of all the [registered HTTP methods][Method Registry].

EDIT: I've currently opened the PR #332 to resolve this issue.

### About Mozilla's documentation
Mozilla's [HTTP request methods][MDN http methods] only contains information about the methods specified in [RFC 7231][] and [RFC 5789][]. While these are the most commonly used ones, they only account for 9 out of the 39 (~23%) currently registered HTTP methods.

### What is a *Registered HTTP Method*?
[RFC 7231][] states the following:

> 4  Request Methods
> 4.1  Overview
>
> ...
> 
> Additional methods, outside the scope of this specification, have
> been standardized for use in HTTP.  All such **methods ought to be
> registered within the "Hypertext Transfer Protocol (HTTP) Method
> Registry" maintained by IANA**, as defined in Section 8.1.
> 
> ...
>
> 8  IANA Considerations
> 8.1  Method Registry
> **The "Hypertext Transfer Protocol (HTTP) Method Registry" defines the
> namespace for the request method token (Section 4)**.  The method
> registry has been created and is now maintained at
> <http://www.iana.org/assignments/http-methods>.

Essentially a registered HTTP method is a method that has gone through the process of being added to [IANA's HTTP Method Registry][Method Registry].

[MDN http methods]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
[RFC 7231]: https://tools.ietf.org/html/rfc7231
[RFC 5789]: https://tools.ietf.org/html/rfc5789
[Method Registry]: https://www.iana.org/assignments/http-methods

yoshuawuyts:
Closes by #332.

Mart-Bogdan:
There should have been open member Other(String) to support further methods, that can be added to IANA later, or ones that can be used privately by applications. Many HTTP servers just use String in place of method

Fishrock123:
I agree

PR #332, opened 2021-02-06
Support all the registered HTTP methods in the `Method` enum

zenekron:
This PR extends the `Method` enum to add all the missing methods from [IANA's HTTP Method Registry][Method Registry].
See the issue #331 . 

Changes:
  - added missing HTTP methods
  - updated `AsRef<str>`, `FromStr` and `Method::is_safe` accordingly
  - modified `Display` to leverage `AsRef<str>`
  - added documentation and links to the appropriate RFC sections
  - added a non-exhausive test that helps catch typos

[Method Registry]: https://www.iana.org/assignments/http-methods

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -5,49 +5,354 @@ use std::str::FromStr;
 
 /// HTTP request methods.
 ///
-/// [Read more](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)
+/// See also [Mozilla's documentation][Mozilla docs], the [RFC7231, Section 4][] and
+/// [IANA's Hypertext Transfer Protocol (HTTP) Method Registry][HTTP Method Registry].
+///
+/// [Mozilla docs]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+/// [RFC7231, Section 4]: https://tools.ietf.org/html/rfc7231#section-4
+/// [HTTP Method Registry]: https://www.iana.org/assignments/http-methods/http-methods.xhtml
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Method {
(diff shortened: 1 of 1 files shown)


jbr:
Out of curiosity, what is the use case for having WebDAV extensions in http-types?

zenekron:
> Out of curiosity, what is the use case for having WebDAV extensions in http-types?

Besides being able to speak WebDAV itself, there are other tools that reuse subsets of these methods for other purposes.

In my case specifically: I'm porting a small Terraform HTTP backend I wrote from warp to tide, but in order to implement state locking I need to be able to handle `LOCK` and `UNLOCK` requests ([link](https://github.com/hashicorp/terraform/blob/2635b3b023379908b7036a5ff9d65ab6021eceee/backend/remote-state/http/backend.go#L49)).

jbr:
Does this mean you'd also want affordances in tide's router for these methods?

zenekron:
> Does this mean you'd also want affordances in tide's router for these methods?

I'm not too sure what you mean by that (not a native English speaker here), but if you're talking about creating methods like `tide::Route::get` and `tide::Route::post` that wrap around `tide::Route::method`, then I think these HTTP methods are uncommon enough that doing so would not bring much benefit.

That being said if you think it's appropriate I can open a PR against tide as well, just let me know.

Fishrock123:
> if you're talking about creating methods like `tide::Route::get` and `tide::Route::post` that wrap around `tide::Route::method`

That is indeed what @jbr meant.

yoshuawuyts:
> I think these HTTP methods are uncommon enough that doing so would not bring much benefit.

I agree with this rationale; I don't think we need to expose shorthands for them, but having them as part of the enum seems reasonable enough.

yoshuawuyts (approved):
Amazing; thank you!

PR #333, opened 2021-02-11
Request: allow Deserialize<'de> for .query()

Fishrock123:
Allows us to avoid extra allocations from querystring parsing.

I don't _think_ this is breaking? But maybe I don't understand the trait inheritance here correctly.

Will also necessitate similar changes in Tide & Surf.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -620,6 +620,8 @@ impl Request {
     /// use http_types::Request;
     /// use std::collections::HashMap;
     ///
+    /// // An owned structure:
+    ///
     /// #[derive(Deserialize)]
     /// # #[serde(crate = "serde_crate")]
     /// struct Index {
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
I remember trying this in the early days of `surf`, and not being able to allocate created all sorts of gnarly lifetime issues. I'm not against this change, but I'd like to see this paired with some extra tests to ensure it in fact does cover all scenarios we want to cover. The risk is that we make this change, and people now need to work around our `serde` methods because they now only cover a subset of uses they did before.

Fishrock123:
I don't follow. As far as i am aware `DeserializeOwned` is more restrictive than `Deserialize<'de>`...?

There is an existing owned test case with deserializing into a hashmap of `Strings`s. This adds a borrowing test case.

Fishrock123:
(See also: https://serde.rs/lifetimes.html)

Fishrock123:
From the aforelinked docs, particularly that last line (emphasis mine):

> ### The `Deserialize<'de`> lifetime
> This lifetime records the constraints on how long data borrowed by this type must be valid.
> 
> Every lifetime of data borrowed by this type must be a bound on the `'de` lifetime of its `Deserialize` impl. If this type borrows data with lifetime `'a`, then `'de` must be constrained to outlive `'a`.
> 
> **If this type does not borrow any data from the `Deserializer`, there are simply no bounds on the `'de` lifetime. Such types automatically implement the `DeserializeOwned` trait.**

Fishrock123:
All signs point to this being fine. Merging.

PR #334, opened 2021-02-11
add a serde feature that's enabled by default

jbr:
in order to call the feature "serde," the crate is renamed to `serde_crate`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["fs", "cookie-secure"]
+default = ["fs", "cookie-secure", "serde"]
 docs = ["unstable"]
 unstable = []
 hyperium_http = ["http"]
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -16,37 +16,41 @@ features = ["docs"]
...
 async_std = ["fs"]
 cookies = ["cookie"]
 cookie-secure = ["cookies", "cookie/secure"]
 fs = ["async-std"]
+serde = ["serde_qs", "serde_crate", "serde_json", "serde_urlencoded"]
 
 [dependencies]
-# features: async_std
-async-std = { version = "1.6.0", optional = true }
+rand = "0.7.3"
```
jbr, line 30: explanation for why unrelated liens changed: I moved the non-optional crates to the top of the list

Fishrock123 (approved):

src/status_code.rs
```diff
@@ -538,82 +536,90 @@ impl StatusCode {
         }
     }
 }
+#[cfg(feature = "serde")]
```
yoshuawuyts, line 540:
```suggestion

#[cfg(feature = "serde")]
```

yoshuawuyts (approved):
:100:

PR #335, opened 2021-02-12
WIP HeaderValue: make Cow<'static, str>

Fishrock123:
This allows HeaderValue to borrow `'static` strings, rather than make allocations.

Unfortunately, we can't specialize `TryFrom` or `From`, so this also adds a new crate-only `from_static_str() -> HeaderValue`.

### DIFF
--- a/src/headers/header_value.rs
+++ b/src/headers/header_value.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::fmt::{self, Debug, Display};
 use std::str::FromStr;
@@ -11,7 +12,7 @@ use crate::Mime;
 /// A header value.
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct HeaderValue {
-    inner: String,
+    inner: Cow<'static, str>,
 }
 
 impl HeaderValue {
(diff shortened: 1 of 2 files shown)


src/headers/header_value.rs
```diff
@@ -11,7 +12,7 @@ use crate::Mime;
 /// A header value.
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct HeaderValue {
-    inner: String,
+    inner: Cow<'static, str>,
```
jbr, line 15: why not `pub struct HeaderValue<'a> { inner: Cow<'a, str> }` and just use `'static` where we need it to be static?

joshtriplett:
Most header names (and many header values) will be short; we might want to use something like `kstring` here, with a small-string optimization. kstring also supports Cow.

src/headers/header_value.rs
joshtriplett: @jbr That'd make the API a little more awkward to use, and would potentially be a backwards-compatibility issue.

jbr:
> Most header names (and many header values) will be short; we might want to use something like `kstring` here, with a small-string optimization. kstring also supports Cow.

[trillium's header name implementation](https://github.com/trillium-rs/trillium/blob/main/http/src/header_name.rs#L19) does something like this, using [smartcow](https://github.com/jbr/smartcow) which is just a cow for [https://docs.rs/smartstring/latest/smartstring/](smartstring). However it might also make sense to use some sort of string interning library because there are only a handful of header names used in reality and they're repeated across requests

PR #340, opened 2021-02-23
Cleanup headers

yoshuawuyts:
Precursor to https://github.com/http-rs/http-types/pull/335, follow-up to #315. This removes all the manual `name`, `value`, and `apply` methods defined on the typed headers in favor of the `Header` trait. This also adds conversions to easily go from `&impl Header` -> `HeaderName` / `HeaderValue`, which allows for more convenient use. Which also enabled us to remove a fair amount of code from the manual `IntoHeaderValues` implementations on various typed headers.

## Example
**before**
```rust
let authz = BasicAuth::new(username, password);

let mut res = Response::new(200);
authz.apply(&mut res);                                         // variant 1, inverted logic
res.insert_header(authz.header_name(), authz.header_value());  // variant 2, clearer but verbose
```
**after**
```rust
let authz = BasicAuth::new(username, password);

let mut res = Response::new(200);
res.insert_header(&authz, &authz); // enabled by this PR
```

## Future directions

This doesn't preclude changing `Header::insert_header` to take one `Header` argument rather than two separate ones. But at least under the current API it significantly simplifies usage patterns and removes lots of duplicates.

### DIFF
--- a/src/auth/authorization.rs
+++ b/src/auth/authorization.rs
@@ -13,7 +13,7 @@ use crate::headers::{Header, HeaderName, HeaderValue, Headers, AUTHORIZATION};
 /// ```
 /// # fn main() -> http_types::Result<()> {
 /// #
-/// use http_types::Response;
+/// use http_types::{headers::Header, Response};
 /// use http_types::auth::{AuthenticationScheme, Authorization};
 ///
 /// let scheme = AuthenticationScheme::Basic;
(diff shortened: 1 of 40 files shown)


yoshuawuyts:
Don't think much here should be controversial -- this is mostly just a big cleanup. Going to go ahead and merge.

PR #342, opened 2021-02-25
Add Body::chain to create a Body from a series of two Body instances

joshtriplett:
This method helps when prepending or appending some data to a Body.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -482,6 +482,46 @@ impl Body {
     pub fn set_mime(&mut self, mime: impl Into<Mime>) {
         self.mime = mime.into();
     }
+
+    /// Create a Body by chaining another Body after this one, consuming both.
+    ///
+    /// If both Body instances have a length, and their sum does not overflow,
+    /// the resulting Body will have a length.
+    ///
+    /// If both Body instances have the same fallback MIME type, the resulting
+    /// Body will have the same fallback MIME type; otherwise, the resulting
+    /// Body will have the fallback MIME type `application/octet-stream`.
+    ///
+    /// # Examples
+    ///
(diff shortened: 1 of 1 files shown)


src/body.rs
```diff
@@ -482,6 +482,46 @@ impl Body {
...
+    /// ```
+    #[cfg(feature = "async-std")]
+    pub fn chain(self, other: Body) -> Self {
+        let mime = if self.mime == other.mime {
+            self.mime.clone()
+        } else {
+            mime::BYTE_STREAM
+        };
+        let length = match (self.length, other.length) {
+            (Some(l1), Some(l2)) => (l1 - self.bytes_read).checked_add(l2 - other.bytes_read),
```
jbr, line 515: bytes_read should always be zero at this point, and if not, that may indicate a larger issue. the body isn't read until the response is handed back to async-h1, at which point we do not expect mutation. additionally, if we've read from both bodies already, the two bodies will be interleaved and be corrupted. We should probably panic if either has a nonzero bytes read, as that's an invalid and irrecoverable state

joshtriplett: @jbr I originally started with that approach, but I *think* this actually works. I can add some more tests to cover this case.

The specific use case that prompted me to not just assert at the top of `chain`: start a process, wrap its Reader in a Body, read some data from the front of that Reader, then prepend some other data in front of it. I *think* that'll work fine here; both `Body` instances will handle returning their remaining bytes.

joshtriplett: I just added a test to make sure this case works correctly.

yoshuawuyts (approved):
neat!

PR #345, opened 2021-03-03
feat: add Error::from_display & from_debug

Fishrock123:
This is handy for errors which are not `Send + Sync + 'static`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -161,6 +161,28 @@ impl Error {
     pub fn type_name(&self) -> Option<&str> {
         self.type_name.as_deref()
     }
+
+    /// Converts anything which implements `Display` into an `http_types::Error`.
+    ///
+    /// This is handy for errors which are not `Send + Sync + 'static` because `std::error::Error` requires `Display`.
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -161,6 +161,16 @@ impl Error {
     pub fn type_name(&self) -> Option<&str> {
         self.type_name.as_deref()
     }
+
+    /// Lossily converts a `std::error::Error` into an `http_types::Error`,
+    /// loosing any information that is not included in the error's `Display` implementation.
```
yoshuawuyts, line 166:
```suggestion
    /// losing any information that is not included in the error's `Display` implementation.
```

yoshuawuyts (approved):
I wonder if `from_error_lossy` might a more descriptive naming, since "lossy" [is only ever used as a suffix](https://doc.rust-lang.org/std/?search=lossy) in the stdlib. But if that might be too confusing, it's fine as-is too.

Fishrock123:
@yoshuawuyts Is this more in line with what you'd hope for? Or is this straying to far from `from_error_lossy`?

yoshuawuyts (approved):

PR #348, opened 2021-03-06
ci: another round of improvements

Fishrock123:
Some things have changed, nightly is being fun again, and some dependency has decided to start throwing a fit when compiling on wasm.


Fishrock123:
needed to make other changes to fix the wasm build

PR #354, opened 2021-04-09
Fixes error messages in Content-Type and Transfer-Encoding negotiation

Tpt:


Fishrock123 (approved):
Thanks for catching this!

PR #355, opened 2021-04-09
Fixes Accept wildcard

Tpt:
It is "*/*" according to https://tools.ietf.org/html/rfc7231#section-5.3.2


Fishrock123:
(is this semver-major?)

Tpt:
> (is this semver-major?)

Not sure, it changes the system behavior but it won't break compilation and fixes a wrong behavior.

PR #359, opened 2021-04-27
fix: append from hyperium translation

Fishrock123:
The opposite direction also uses hyperium's append, but (I think) this was written before that same functionality existed in http-types.

This is somewhat important for my use and I intend to make a 2.11.1 release with this.


jbr (approved):
Seems reasonable. It overwrote previously?

Fishrock123:
Err, actually, it _theoretically_ overwrote previously but it actually just unconditionally panic-ed on any new header, since we're unwrapping the option from https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert ...

I'll update the commit message to clarify that this _really_ did not work before.

PR #364, opened 2021-05-20
Implement `Status` for `Result<T, http_types::Error>`

pbzweihander:
It will be very convenient if I can use `Status` trait for `Result<T, http_types::Error>` to replace StatusCode.

For example, (in tide)

```rust
let body = request.body_json().await.map_err(|mut e| {e.set_status(StatusCode::BadRequest); e})?;
let data = some_function(body).status(StatusCode::NotFound)?;
```

...will become:

```rust
let body = request.body_json().await.status(StatusCode::BadRequest)?;
let data = some_function(body).status(StatusCode::NotFound)?;
```

### DIFF
--- a/src/cache/age.rs
+++ b/src/cache/age.rs
@@ -62,7 +62,7 @@ impl Age {
         // entry. We want the last entry.
         let header = headers.iter().last().unwrap();
 
-        let num: u64 = header.as_str().parse().status(400)?;
+        let num: u64 = header.as_str().parse::<u64>().status(400)?;
         let dur = Duration::from_secs_f64(num as f64);
 
         Ok(Some(Self { dur }))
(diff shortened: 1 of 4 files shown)


Fishrock123 (approved):

src/cache/age.rs
```diff
@@ -62,7 +62,7 @@ impl Age {
         // entry. We want the last entry.
         let header = headers.iter().last().unwrap();
 
-        let num: u64 = header.as_str().parse().status(400)?;
+        let num: u64 = header.as_str().parse::<u64>().status(400)?;
```
Fishrock123, line 65: @pbzweihander Did these suddenly become required?

Fishrock123 (commented):
Just wondering if this need to be semver-major.

src/cache/age.rs
pbzweihander: Yes. But I don't know exactly why...

Fishrock123: I think it's because `.status()` uses `TryInto` inference but `?` also uses similar inference.

I still think this is probably a good idea (at least until `http-types::Error` can impl `StdError`), but it means this would have to happen in 3.0.

PR #368, opened 2021-06-22
Update serde-qs

pbzweihander:
`serde-qs` is updated to `0.8`, so it would be great if `http-types` use latest `serde-qs`.

For my case, I needed [https://github.com/samscott89/serde_qs/commit/b7b2520107ecfca77520151c6dc3002442bf0f0b](this commit) to encode query correctly.

Currently, I'm using this workaround:

```
pub fn set_query<Req>(req: Req, query: &impl serde::Serialize) -> http_types::Result<Req>
where
    Req: Into<http_types::Request> + From<http_types::Request>,
{
    let mut req: Request = req.into();
    let query = serde_qs::to_string(query)
        .map_err(|e| http_types::Error::from_str(StatusCode::BadRequest, format!("{}", e)))?;
    req.url_mut().set_query(Some(&query));
    Ok(req.into())
}
```


jbr (approved):

jbr:
Thanks!

PR #369, opened 2021-06-22
Expose reference of inner anyhow error

pbzweihander:
It would be great if `http_types::Error` exposes the reference of inner `anyhow::Error`.

Especially, when using `backtrace` feature of `anyhow` in stable rust, I cannot get backtraces without losing status and `type_name` information. But if I can get the reference of `anyhow::Error`, I can without losing information.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -238,3 +238,9 @@ impl From<Error> for Box<dyn StdError + 'static> {
         Box::<dyn StdError + Send + Sync>::from(error.error)
     }
 }
+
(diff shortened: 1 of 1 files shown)


jbr:
Could this be an `impl AsRef<anyhow::Error> for http_types::Error`?

pbzweihander:
@jbr I updated the PR with `impl AsRef<anyhow::Error> for http_types::Error`!

jbr:
Thanks!

PR #375, opened 2021-07-30
address three bugs in forwarded header parsing

### LINKED ISSUE #374: Error in parser for the "Forwarded" header
The `proxies::forwarded::Forwarded::parse` method currently (i'm using version 2.11.1) assumes, that all `for=` items are at the beginning of the forwarding header. As far as i can see from https://datatracker.ietf.org/doc/html/rfc7239#page-5 the RFC does not specify a specific position for the `for=` forwarded-pair.

jbr:
Previously,
* parsing forwarded headers required the `for` section to be first (#374)
* checking short strings against a prefix that is longer than it with `starts_with_ignore_case` would panic (only used in forwarded currently)
* parse_utils::parse_token (`pub(crate)`) token did not successfully parse 1-character tokens due to an off-by-one error

This can be fairly easily backported to the 2.x line if there's interest

closes #374

### DIFF
--- a/src/parse_utils.rs
+++ b/src/parse_utils.rs
@@ -5,7 +5,7 @@ pub(crate) fn parse_token(input: &str) -> (Option<&str>, &str) {
     let mut end_of_token = 0;
     for (i, c) in input.char_indices() {
         if tchar(c) {
-            end_of_token = i;
+            end_of_token = i + 1;
         } else {
             break;
         }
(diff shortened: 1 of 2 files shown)


Fishrock123 (approved):
Would be nice to backport

jbr:
do we have a branch or tag for 2.x?

Fishrock123:
@jbr https://github.com/http-rs/http-types/tree/v2.x.x-backports

PR #382, opened 2021-09-16
Fix correct default weight in media_type_propose

### LINKED ISSUE #362: Accept header sort does not use the correct default weight value
[RFC 7231 states](https://tools.ietf.org/html/rfc7231#section-5.3.1) that the default value for the weight is 1. However the `accept` sort method ranks the values without weights after the ones with a weight lower than 1, even if the should be sorted first.

u5surf:
Fixes #362 
https://tools.ietf.org/html/rfc7231#section-5.3.1
refers that default weight value is 1.0. So, I fixed if weight in media_type_propose is None, it applies default value.

### DIFF
--- a/src/content/accept.rs
+++ b/src/content/accept.rs
@@ -362,9 +362,9 @@ mod test {
         let mut accept = Accept::from_headers(headers)?.unwrap();
         accept.sort();
         let mut accept = accept.iter();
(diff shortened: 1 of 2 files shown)


yoshuawuyts (approved):
Hi, thanks for opening this PR. I checked the RFC you linked, and indeed - it should default to 1.0. Thanks so much for fixing this!

PR #386, opened 2021-10-30
Change hard-coded headernames to lowercase

halvko:
The change at line 114 in `src/security/mod.rs` corrects a logic error, the rest is at best a small optimization and at worst a style choice

### DIFF
--- a/src/security/csp.rs
+++ b/src/security/csp.rs
@@ -357,9 +357,9 @@ impl ContentSecurityPolicy {
     /// Sets the `Content-Security-Policy` (CSP) HTTP header to prevent cross-site injections
     pub fn apply(&mut self, mut headers: impl AsMut<Headers>) {
         let name = if self.report_only_flag {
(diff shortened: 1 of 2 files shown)


yoshuawuyts (changes requested):
I'm not sure I understand what issue this solves? Header names should be case-agnostic. And since we're only ever encoding values here, if an error occurs it seems like the issue would be in a parser, not in any of the encoders?

I feel like I'm missing something?

halvko:
As I understand it, headers are not case agnostic for HTTP/2 connections “However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2” (https://httpwg.org/specs/rfc7540.html#HttpHeaders).

halvko:
And the documentation for ‘headername::from_lowercase_str’ specifically states that the input is expected to be lowercase

halvko:
I'll close this as it's old and no interest was shown in merging it

Issue #381, opened 2021-09-01
`Body::set_mime` should take `Option<impl Into<Mime>>`


yoshuawuyts:
It's currently impossible to unset the mime associated with the body. We should make it so we can unset the mime passed to the body.

jbr:
I believe this doesn't work because the compiler can't infer the impl generic for None. We could have a Body::remove_mime, maybe?

halvko:
@yoshuawuyts The code in the PR is finished (and correct), but I've shown in the doc comment of the function why it (as @jbr hints at) a somewhat ugly solution. As I write in the commit message, there are a couple of alternatives: Better documentation, or the introduction of a function.

There is also the possibility I've misunderstood you and you want `Body::mime` to be an `Option`, but I'm not familiar enough with the code base to know if that would make sense.

yoshuawuyts:
> There is also the possibility I've misunderstood you and you want `Body::mime` to be an `Option`, but I'm not familiar enough with the code base to know if that would make sense.

Yeah, that's what I was thinking 😅 . I ran into this when trying to write a converter for [`vcr-cassette`](https://github.com/http-rs/vcr-cassette) where we couldn't express a body without a mime type. Good defaults are important, but having a way to actually unset the mime type seems like something we should support.

halvko:
@yoshuawuyts #387 should do this correctly now, and doesn't add any _new_ warnings

halvko:
@yoshuawuyts As #387 was merged, I think this should be closed

PR #387, opened 2021-10-30
Make 'mime' field in Body optional

halvko:
This does not seem like the ideal way to do this since Option have to be
parameterized with a type, see the introduced example code for specific
example.

I see a couple of alternatives:
 - Add a method for unsetting the mime type
 - Clearly document what mime type is the "no mime type". This should
   possibly be part of an example above set_mime

The first as a semver minor change, and the second is a semver patch
change.

A solution to #381

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -483,11 +483,11 @@ impl Body {
     /// # Examples
     /// ```
     /// use http_types::Body;
-    /// use http_types::mime::self;
+    /// use http_types::mime;
     ///
-    /// let mut body = Body::default();
-    /// body.set_mime(Some(mime::CSS));
-    /// assert_eq!(*body.mime(), mime::CSS);
+    /// let mut body = Body::empty();
+    /// body.set_mime(mime::CSS);
+    /// assert_eq!(body.mime(), Some(&mime::CSS));
     /// ```
     pub fn set_mime(&mut self, mime: impl Into<Mime>) {
         self.mime = Some(mime.into());
@@ -500,7 +500,7 @@ impl Body {
     /// use http_types::Body;
     /// use http_types::mime;
     ///
-    /// let mut body = Body::default();
+    /// let mut body = Body::empty();
     /// assert!(body.mime().is_some());
     ///
     /// body.unset_mime();
--- a/src/body.rs
+++ b/src/body.rs
@@ -486,29 +486,14 @@ impl Body {
     /// use http_types::mime;
     ///
     /// let mut body = Body::empty();
-    /// body.set_mime(mime::CSS);
+    /// body.set_mime(Some(mime::CSS));
     /// assert_eq!(body.mime(), Some(&mime::CSS));
+    ///
+    /// body.set_mime(None);
+    /// assert_eq!(body.mime(), None);
     /// ```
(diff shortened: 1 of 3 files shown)


halvko:
Oops, forgot to commit ^^

src/body.rs
```diff
@@ -474,13 +474,41 @@ impl Body {
...
+    ///
+    /// let mut body = Body::default();
+    /// assert!(body.mime().is_some());
+    ///
+    /// body.unset_mime();
+    /// assert!(body.mime().is_none());
+    /// ```
+    ///
+    pub fn unset_mime(&mut self) {
+        self.mime = None
```
yoshuawuyts, line 511: I was hoping we could let `set_mime` take an `Option`.

yoshuawuyts (commented):
Feels like we're getting closer!

src/body.rs
halvko: If we want to accept an `Option<impl Into<Mime>>`, all uses of `None` in argument position would have to be disambiguated for the compiler:
```rust
body.set_mime(Option::<Mime>::None);
```
This (I think) doesn't look great, and if the user does it wrong, the compiler doesn't give a great error message (because it can't infer what might be a reasonable type for T):

```rust
error[E0283]: type annotations needed
 --> src/body.rs:489:6
  |
8 | body.set_mime(None);
  |      ^^^^^^^^ cannot infer type for type parameter `T` declared on the associated function `set_mime`
  |
  = note: cannot satisfy `_: Into<Mime>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0283`.
```

If we want the generic Into<Mime>, I think the best alternative would be an unset function. An alternative (still keeping Into would be changing the signature to

```rust
pub fn set_mime(&mut self, mime: impl Into<Option<Mime>>);
```

but I don't like this for
1. A performance reason: It introduces an otherwise unnecessary branch in set_mime (a bit of code I believe could be called quite often), and is unlikely to be inlined as it crosses (at least) one create boundary (this would have to be benchmarked though, and could probably be solved with `#[inline(always)]`),
2. A ecosystem breakage reason: It would break any custom Mime implementation both in binary and library code for (in my opinion) not much gain, and
3. A style reason: I think it seems weird implementing `Into` for an `Option`, and using the error type of `TryInto` as unsetting the mime could lead to bad debugging experiences (tracing through calls until you find something just eating the error).

I'll add that I think it would be a fine option to change the signature to:
```rust
pub fn set_mime(&mut self, mime: Option<Mime>)
```
This would solve all problems with disambiguation, and ensures that "bad" Into implementations panic in the callers code, but I'm not sure how much developer experience would suffer (or if at all).

yoshuawuyts: > I'll add that I think it would be a fine option to change the signature to:
> 
> ```rust
> pub fn set_mime(&mut self, mime: Option<Mime>)
> ```
> 
> This would solve all problems with disambiguation, and ensures that "bad" Into implementations panic in the callers code, but I'm not sure how much developer experience would suffer (or if at all).

That indeed seems like the better option. Happy for us to with that!

halvko:
@yoshuawuyts Does it seem allright?

yoshuawuyts (approved):
:100:

Issue #371, opened 2021-07-10
UTF-8 mime-type constants don't work well with browsers' `Accept` header


lo48576:
Some of predefined text mime-types (such as `mime::HTML` and `mime::XML`) have `;charset=utf-8` parameter, but browsers' default `Accept` header does not.
(See [List of default Accept values - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values) for browsers' defaults.)

When servers (written in Rust) uses `mime::{HTML, XML, ...}` to represent available content types, `content::Accept::negotiate()` does not work well with browsers' default `Accept`, and it fails with "No suitable Content-Type found" error.

```rust
let mime_html = "text/html".parse::<Mime>()?;
let mime_xhtml = "application/xhtml+xml".parse::<Mime>()?;

// This is simple version of browser's `Accept` value.
let mut browser_accept = Accept::new();
browser_accept.push(MediaTypeProposal::new(mime_html, None)?);
browser_accept.push(MediaTypeProposal::new(mime_xhtml, None)?);

// This is server's default.
let acceptable = &[mime::HTML];

let res = Response::new(200);
let content_type = browser_accept.negotiate(acceptable);

// I expected this to success, but fails!
assert!(
    content_type.is_ok(),
    "server is expected to return HTML content"
);
```

So, the questions are:

* Should difference of parameters' existence cause negotiation to fail?
* Is UTF-8 MIME constants (`mime::{JAVASCRIPT, CSS, HTML, PLAIN, XML}` as of http-types 2.11.1) intended to be used by servers?
    + If so, how they make negotiation succeed when requests are from browsers?
    + If not, are servers recommended to execute `"text/html".parse::<Mime>().unwrap()` as they need, or are there more better ways?

Fishrock123:
@yoshuawuyts

yoshuawuyts:
Hey, thanks for filing this. That's a really interesting question! I think your breakdown of the two questions is exactly right, and we should seek to answer those.

### Should difference of parameters' existence cause negotiation to fail?

[RFC7231:3.1.1](https://httpwg.org/specs/rfc7231.html#rfc.section.3.1.1.1) has the following to say:
> The presence or absence of a parameter might be significant to the processing of a media-type, depending on its definition within the media type registry.

[...]

PR #389, opened 2021-11-06
Change mime negotiation

halvko:
I've extended Mime with a subset operation, which describes being/accepting a subset of another Mime. This solves the problem described in #371.


Fishrock123:
I like it.

yoshuawuyts (approved):

Issue #361, opened 2021-05-04
Accept header sort is not stable leading to unstable negotiation result


Tpt:
The following test passes:

```rust
#[test]
fn negotiate_tie() -> crate::Result<()> {
        let mut accept = Accept::new();
        accept.push(MediaTypeProposal::new(mime::HTML, Some(0.4))?);
        accept.push(MediaTypeProposal::new(mime::XML, Some(0.4))?);

        assert_eq!(accept.negotiate(&[mime::HTML, mime::XML])?, mime::XML);
        assert_eq!(accept.negotiate(&[mime::XML, mime::HTML])?, mime::HTML);
        Ok(())
}
```

`XML` (or `HTML`?) should be picked the two times.

PR #390, opened 2021-11-20
Fix Accept header sort order #361

u5surf:
Fixes: #361 

* Implements Clone for Accept in order to copy it to tmporary variable and evaluate with it inside negotiate method, so that self cannot be sorted each negotiate is called.

* Add a test which evaluates negotiation properly the header which has same weight.

### DIFF
--- a/src/content/accept.rs
+++ b/src/content/accept.rs
@@ -48,6 +48,7 @@ use std::slice;
 /// #
 /// # Ok(()) }
 /// ```
+#[derive(Clone)]
 pub struct Accept {
     wildcard: bool,
     entries: Vec<MediaTypeProposal>,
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Thanks for opening this PR! On reviewing the spec, you're indeed right that our current implementation is not compliant - but this PR doesn't implement the ordering dictated by the spec either.

I've written an analysis of the issues we have in our content negotiation in https://github.com/http-rs/http-types/issues/361#issuecomment-1008872107, and my hope is that we can use that to implement a spec-compliant fix.

I'm going to close this PR for now since it doesn't match the spec. But I'd be keen to review any follow-up PRs which help bring our content-negotiation implementation according to spec. Thanks!

PR #395, opened 2022-02-04
refactor: overhaul errors

Fishrock123:
Overhauls errors to allow for 3 types of errors:

- Concrete errors (enum) from http-types itself
- A dynamic error type for response handlers (almost identical to how `Error` was previously)
- A combined type for handling request+response for a client

There is some exposition for this in Zulip: https://http-rs.zulipchat.com/#narrow/stream/261414-http-types/topic/New.20error.20types.20.26.20result.20apis/near/254273467

Basically this solves two problems:
- Errors from within http-types/tide/surf ware currently unable to be matched upon except by string search.
- Surf's error return type not being StdError is a massive pain point.
    - Actually this is a pain point for certain other operations to, such as creating global headers during startup.
    
The first problem is pretty simple to solve but was quite laborious. It meant changing all those `bail!("some string error")` and what not to be concrete enum variants to some degree of specificity. I have kept some errors grouped together where the string output could mean something to anyone debugging a problem but is unlikely to be something which someone would reasonably want to check during program runtime. This is to reduce the amount of labor doing this but also to cap the enum variants and prevent the number of variants from becoming unmanageable.

The second problem is tricky and weird. My solution is `RequestError`. This allows dynamic errors from surf middleware but also allows us to have concrete errors there. Some indirection is required for `ResponseError` due to `StdError`'s conflict with `anyhow` (The problem is a conflict with `From<T> for T`, which I will never not be profusely annoyed at).

-----

There are some potential issues here / open questions:

- Does this actually work well with Tide and Surf?
    - Basically, those libraries will also need their own concrete error types, and one of the varients would be `HTTP(http_types::Error)` or such, which... isn't really correct.
    - We could solve that by removing the central error enum here and only having the specific sub error types `HeaderError`, `BodyError`, etc. Would need to make a couple extra types.
    - We could also reinterpret the main http-types error enum and translate it directly to surf and tide ones.
    - I have a slightly worry about how this would work between tide and surf, i.e. Tide would have to turn a Surf concrete error into a dynamic error unless it shared the same error enum for all cases even from those libraries.
- Should `Error::ArgTryIntoError(Box<dyn std::error::Error + Send + Sync + 'static>)` exist? Would returning `Result<Result<T>>` in the necessary places be better? (@yoshuawuyts this is what I was asking on Zulip)
- Are the `Send + Sync + 'static` requirements correct or a problem?


Fishrock123:
Hmph. The CI duly notes that this makes interop with hyper http more difficult (obviously).

Cargo.toml
```diff
@@ -50,6 +50,8 @@ serde_json = { version = "1.0.51", optional = true }
 serde_crate = { version = "1.0.106", features = ["derive"], optional = true, package = "serde" }
 serde_urlencoded = { version = "0.7.0", optional = true}
 serde_qs = { version = "0.8.3", optional = true }
+miette = "2.2.0"
```
zkat, line 53: why this version?

Fishrock123: ah, so I actually forgot about this until I submitted the PR haha because I originally drafted parts of it in like, September 2021.

I should probably remove it for now and then re-add it in a follow-up PR, but at least this proved that it didn't break anything from this patch

Fishrock123:
@yoshuawuyts Hey I'm going to merge this tomorrow if there is no objection

Fishrock123:
Made one additional change to this... all the enums are `#[non_exhaustive]` now so we can add more cases in minors where it fits well

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[error("Header error: {}", .0)]
+    Header(HeaderError),
+
+    #[cfg(feature = "hyperium_http")]
+    #[error("URL Parse error: {}", .0)]
+    URLParse(#[from] url::ParseError),
+
+    #[cfg(feature = "hyperium_http")]
+    #[error("Hyperium HTTP error: {}", .0)]
+    HyperiumHttp(#[from] http::Error),
```
joshtriplett, line 42: Could we just bundle this into a case for third-party library errors, and support downcasting? That would avoid exposing third-party types in our API, which could introduce breakage if those libraries bump versions.

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[error("HTTP Method was unrecognized: {}", .0)]
+    HttpMethodUnrecognized(String),
+
+    #[error("I/O error: {}", .0)]
+    IO(#[from] std::io::Error),
+
+    #[error("Invalid Status Code: {}", .0)]
+    StatusCodeInvalid(u16),
+
+    #[error("Querystring deserialization error: {}", .0)]
```
joshtriplett, line 25:
```suggestion
    #[error("Query string deserialization error: {}", .0)]
```

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    QueryDeserialize(serde_qs::Error),
+    #[error("Querystring serialization error: {}", .0)]
```
joshtriplett, line 27:
```suggestion
    #[error("Query string serialization error: {}", .0)]
```

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
+use miette::Diagnostic;
+use thiserror::Error as ThisError;
+
+use crate::auth::AuthenticationScheme;
+use crate::StatusCode;
+
+/// Error kind for http-types
+#[derive(Debug, Diagnostic, ThisError)]
```
joshtriplett, line 8: Very minor nit, but I think this code would be marginally easier to understand if it just used `thiserror::Error` here.

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[error("WWW-Authenticate header was invalid: {}", .0)]
+    WWWAuthenticateInvalid(&'static str),
+    #[error("Content-Location header url was invalid: {:?}", .0)]
+    ContentLocationInvalidUrl(url::ParseError),
+    #[error("Content-Location header base url was invalid: {:?}", .0)]
+    ContentLocationInvalidBaseUrl(Box<dyn std::fmt::Debug + Send + Sync + 'static>),
+    #[error("Expect header was malformed.")]
+    ExpectInvalid,
+    #[error("Timing-Allow-Origin header was invalid: {:?}", .0)]
+    StrictTransportSecurityInvalid(&'static str),
```
joshtriplett, line 143: Is this error message mismatched?

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[error("Accept-Encoding header was invalid: {}", .0)]
+    AcceptEncodingInvalidEncoding(EncodingError),
+    #[error(
+        "No suitable Content-Encoding header Encoding found during Accept-Encoding negotiation"
+    )]
+    AcceptEncodingUnnegotiable,
+    #[error("ETag header was invalid")]
+    ETagInvalid,
+    #[error("Age header was invalid: length out of bounds (unsized 64 bit integer)")]
+    AgeInvalid,
```
joshtriplett, line 127: In general, do we want to have one of these variants for every header, or do we want to have one variant that has a header name and a message?

src/errors/error_kind.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[error("HTTP Date-Time invalid hours")]
+    HourInvalid,
+    #[error("HTTP Date-Time invalid day")]
+    DayInvalid,
+    #[error("HTTP Date-Time invalid month")]
+    MonthInvalid,
+    #[error("HTTP Date-Time invalid year")]
+    YearInvalid,
+    #[error("HTTP Date-Time invalid week-day")]
+    WeekdayInvalid,
```
joshtriplett, line 210: Is anyone going to use individual variants like this, rather than a more general parse error? When would someone want to treat these programmatically differently, rather than just showing a different error?

joshtriplett:
Posted a few bits of feedback, but generally this looks great. Thanks for the substantial amount of work here.

src/errors/error_kind.rs
Fishrock123: Another question would be naming... how do we feel about `<header name>Invalid<optional detailed invalid area>`?

src/errors/error_kind.rs
Fishrock123: Hmm. The existing code just made all these errors and I kinda replicated it... I don't know if it would be useful... if not these could probably be `&'static str`s...

src/errors/error_kind.rs
Fishrock123: @joshtriplett What would that look like? Specifically the "bundle this into a case for third-party library errors" part of that

Fishrock123:
@yoshuawuyts if you are going to write a bunch of typed headers could you please take a look at this first?

yoshuawuyts:
@Fishrock123 thanks so much for the ping! - we were actually running into very similar issues around error handling while looking to integrate `http-types` into the Azure SDK (https://github.com/Azure/azure-sdk-for-rust/issues/848), and I'm fairly comfortable saying that this seems like the right direction to go.

I still need to look at this line-by-line, but want to at least respond that I agree that we should land this before any further work on typed headers.

rbtcollins:
I haven't read the diff, but just a quick though since future API breakage seems to be a strong concern here. Consider making all the concrete enums nonexhaustive. That will permit expanding them in future without API breakage - everyone will need to explicitly handle unexpected errors and not assume the error space will expand. (See e.g. filesystem errors in the stdlib for the pain of *not* doing this in advance).

Fishrock123:
> I haven't read the diff, but just a quick though since future API breakage seems to be a strong concern here. Consider making all the concrete enums nonexhaustive. That will permit expanding them in future without API breakage - everyone will need to explicitly handle unexpected errors and not assume the error space will expand. (See e.g. filesystem errors in the stdlib for the pain of _not_ doing this in advance).

Yes, it will be a large api change. I think that’s fine since no one _)really_ likes the current api anyway.

Also, yes I already did mark them all as `#[non_exhaustive]`:
https://github.com/http-rs/http-types/pull/395/files#diff-87f966704ebddb378e447e13ddf9ee203cd9d31bb502338a5f3ed3478df41588R9

OneOfOne:
I'm maintaining a hard fork of this repo that I plan to keep up to date, please resubmit this PR at https://github.com/OneOfOne/http-types-rs

PR #397, opened 2022-03-23
Update infer to 0.7.0

janbaudisch:
The used `infer` version (`0.2.3`) is quite old. The mime type (now being a string slice) is retrieved through a `const fn` instead of a field.


yoshuawuyts (approved):
Looks good, thanks!

PR #396, opened 2022-02-21
Update rand dependency because of RUSTSEC-2021-0023

jhorstmann:
There is an open rustsec issue ([RUSTSEC-2021-0023](https://rustsec.org/advisories/RUSTSEC-2021-0023.html)) about a transitive dependency of http-types 2.12.0. On the main branch this seems to be fixed by migrating to `fastrand` but I think this might also warrant a maintenance release.

PR #399, opened 2022-03-29
Fix all failing clippy lints for backports branch

arlyon:
Hi

I am interested in getting #396 merged and noticed it is blocked due to failing clippy lints on this branch. I have simply run through and fixed them.

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -507,7 +507,7 @@ impl AsyncRead for Body {
         cx: &mut Context<'_>,
         buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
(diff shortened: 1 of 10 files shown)


arlyon:
There seems to be a test that is (consistently) broken on this branch as well, having failed both CI runs: https://github.com/http-rs/http-types/runs/5734932654?check_suite_focus=true

arlyon:
More investigation: tests pass on rust version `1.59.0`, fail on version `1.61.0-nightly`.

Fishrock123:
In the future, you may want to run `cargo clippy --workspace --all-targets` to catch _everything_.

PR #510, opened 2022-09-13
fix backtrace function

mu2019:


jbr (approved):
LGTM, thanks! We probably should at least compile http-types with this feature enabled on CI, either as part of this PR or a subsequent one

PR #516, opened 2023-02-17
updated cookie to 0.17.0

06chaynes:
This bumps cookie to 0.17.0


simonsan (approved):

PureWhiteWu:
Hi, seems that the latest version of cookie is 0.18, can this be bumped to 0.18 and release a new version?

The dependency of cookie 0.14 is not maintained anymore and can lead to compile fail, see: https://github.com/RustCrypto/traits/issues/1609

PR #521, opened 2024-01-08
Updated dependencies

AlexSherbinin:
Updated dependencies, added rust-toolchain.toml file with nightly channel described.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "http-types"
-version = "3.0.0"
+version = "4.0.0"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/http-rs/http-types"
 documentation = "https://docs.rs/http-types"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,31 +27,31 @@ fs = ["async-std"]
 serde = ["serde_qs", "serde_crate", "serde_json", "serde_urlencoded", "url/serde"]
 
 [dependencies]
-fastrand = "1.4.0"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -27,31 +27,31 @@ fs = ["async-std"]
...
+url = "2.5.0"
+anyhow = "1.0.79"
 
 # features: async_std
-async-std = { version = "1.6.0", optional = true }
+async-std = { version = "1.12.0", optional = true }
 
 # features: hyperium/http
-http = { version = "0.2.0", optional = true }
+http = { version = "1.0.0", optional = true }
```
Fishrock123: This is a breaking change requiring a major version

jakubadamw: The newest version on crates.io is 2.12.0, though, so the 3.0.0 in Cargo.toml was fine. 🙂

Fishrock123: it mostly means I will not commit to merging this change

S0c5:
@Fishrock123 is there an interest in merging this PR? We're facing issues with nested dependencies that are outdated and are breaking our apps

Fishrock123:
I will take a dependency upgrade that does not cause a major version (is not a breaking change for exposed interfaces)

I will take an update to tokio/hyper and request/curl only if those updates include it under a separate feature flag, with the dependency version in the flag, which was the big mistake here all along.

in more clear terms, the `hyperium_http` feature should have been `hyperium_http_02` (the `0.2` version), so for hyper http `1.0` it should be `hyperium_http_1`

PR #522, opened 2024-01-31
use 64 bit trace id (for wasm-unknown-emscripten target)

StuartHarris:
Use a 64 bit `trace_id` (instead of 128 bit) so that it can be compiled for `wasm32-unknown-emscripten`


OneOfOne:
I'm starting to manage an active fork of this repo, if you can submit this pr to https://github.com/OneOfOne/http-types-rs and make sure it passes tests, I'll publish it.

